Issue #2, opened 2019-06-28
Is Priority hop-by-hop?


LPardue:
The section `Coexistence with HTTP/2 Priorities` makes me wonder if the header should be treated as hop-by-hop, because it relates to the transfer properties of a particular connection. Naively passing through the header in either direction may cause problems?

kazuho:
I think it needs to be end-to-end.

The values transmitted using the Priority field is how the document that affects the client. That property does not change as it traverses through the intermediaries. Therefore end-to-end. The end-to-end nature is also a prerequisite for an intermediary caching and utilizing the value of the header field.

OTOH, how the header is utilized is a hop-by-hop issue. "Coexistence with HTTP/2 Priorities" discusses that, because HTTP/2 can prioritize responses thanks to the streams.

> Naively passing through the header in either direction may cause problems?

I do not think so. The coexistence section states that "an intermediary SHOULD set the settings parameter for a connection it establishes when and only when all the requests to be sent over that connection originates from a client that utilizes this header-based prioritization scheme".

This means that only a HTTP/2 connection that only carries requests from one client (which either uses Priority headers or not) would have the settings parameter set. In such case, a server respects the Priority header of the requests.

When a HTTP/2 connection contains a mix of requests from client sending Priority headers and those do not, the settings parameter is _not_ set, and the server will use the legacy HTTP/2 prioritization scheme.

I think this "respect the header-based priorities only when all the requests on the connection are capable of using it" approach would work fine regardless of the HTTP versions that intermediaries would support.

kazuho:
@LPardue I've created #4 that clarifies my comment above in Considerations. Please let me know what you think.

PR #3, opened 2019-06-28
Editorial pass and author update

LPardue:
My first intent was to fix typos but then I got carried away. I can split into two PRs if the changes are too contentious.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -17,6 +17,11 @@ author:
     name: Kazuho Oku
     org: Fastly
     email: kazuhooku@gmail.com
+  -
+    ins: L. Pardue
+    name: Lucas Pardue
+    org: Cloudflare
+    email: lucaspardue.24.7@gmail.com
 
 normative:
   RFC2119:
@@ -39,28 +44,29 @@ informative:
 --- abstract
 
 This document describes the Priority HTTP header field.  This header field can
-be used by the endpoints to specify the absolute precedence of a HTTP response
-in an HTTP version-independent way.
+be used by endpoints to specify the absolute precedence of an HTTP response
+in an HTTP-version-independent way.
 
 --- middle
 
 # Introduction
 
-It is often the case for the processing of an HTTP response to depend on the
-retrieval of another HTTP response.
-
-As an example, visual rendering of an HTML document could be blocked by the
-retrieval of a CSS file that the document refers to. Inline images is another
-example, though the difference is that they typically do not block rendering.
-Instead, they are rendered progressively as the chunks of the images arrive.
+It is common for an HTTP resource representation to have relationships to one or
+more other resources. Clients will often discover these relationships while
+processing a retrieved representation, leading to further retrieval requests.
+Meanwhile, the nature of the relationship determines whether the client is
+blocked from continuing to process locally available resources. For example,
+visual rendering of an HTML document could be blocked by the retrieval of a CSS
+file that the document refers to. In contrast, inline images do not block
+rendering and get drawn progressively as the chunks of the images arrive.
 
 To provide meaningful representation of a document at the earliest moment, it is
 important for a HTTP server to prioritize the HTTP responses, or the chunks of
 those HTTP responses, that it sends.
 
-HTTP/2 ({{RFC7540}}) provides such prioritization scheme. A client sends a
-series of PRIORITY frames to communicate to the server a “priority tree;” a tree
-that represents the client's preferred ordering and weighted distribution of the
+HTTP/2 ({{RFC7540}}) provides such a prioritization scheme. A client sends a
+series of PRIORITY frames to communicate to the server a “priority tree”; this
+represents the client's preferred ordering and weighted distribution of the
 bandwidth among the HTTP responses.  However, the design has shortcomings:
 
 * Its complexity has led to varying levels of support by the HTTP/2 client and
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -116,22 +123,23 @@ Argument syntax:
...
 * `non-blocking` indicates that the response does not block the processing of
-  the document even though the response is being used or referred by the
+  the document even though the response is being used or referred to by the
   document.
 
-When the Priority header field or the `urgency` directive do not appear in the
-request, the server SHOULD act as if urgency level of `document` was specified.
+The default value of the `urgency` directive is `document`. This SHOULD apply if
+the directive is omitted, if the Priority header is missing, or if the
+server ignores the Priority header field.
```
kazuho: Could you please clarify the intent of stating "or if the server ignores the Priority header field"?

I am not sure what it suggests, and is confusing.

draft-kazuho-httpbis-priority.md
```diff
@@ -153,31 +161,31 @@ progressive = "yes" / "no"
 This boolean directive indicates if a response can be processed progressively,
 i.e. provide some meaningful output as chunks of the response arrive.
 
-When the Priority header field or the `progressive` directive do not appear in
-the request, the server SHOULD act as if a `progress` directive with an argument
-of `no` has specified.
+The default value of the `progressive` directive is `no`. This SHOULD apply if
+the directive is omitted, if the Priority header is missing, or if the
+server ignores the Priority header field.
```
kazuho: Here too.

draft-kazuho-httpbis-priority.md
```diff
@@ -94,12 +100,13 @@ interpreted as described in [RFC2119].
 
 # The Priority HTTP Header Field
 
-The Priority HTTP header field is used by a client as part of an HTTP request
-to specify the priority of the response.  A server can also use the header field
-as part of an HTTP request to communicate the correct or amended precedence to
-an intermediary.
+The Priority HTTP header field may appear in requests and responses. A client
```
kazuho: How about s/may/can/? "May" is a RFC2119 word.

kazuho (commented):
This is much much better. Thank you very much for the PR.

Looks good to me aside from the three issues below.

draft-kazuho-httpbis-priority.md
kazuho: PS. Now I realize that this might be due to the provision we have in the "Coexistence with HTTP/2" section?

Assuming that is true, and through the discussion in #2, I think we might want to downgrade the "Coexistence ~" section to one of considerations, and remove this "if ignores..." provision for simplicity. Using a boolean settings parameter to disable header-based prioritization is just one way of achieving coexistence, and I think there is room to explore other designs.

WDYT?

LPardue: While editing from start to finish i built up more understanding of the spec. However, I think I'm still a little confused about exactly when the Priority header can appear on requests r responses. The reason being that we state the header is "HTTP version independent". Can this header only appear on requests/responses that are made over HTTP/2 and HTTP/3? What happens if the header is present but the setting is not advertised? Is the priority ignored in favor of frame-based prioritization? Conversely, we state that a default priority applies if the header is missing, but that is only meaningful if the server was expecting to see the header; this leads me to believe that it relies upon negotiation via Settings.

On my first reading it seemed feasible that a HTTP/1.x client could add the Priority header to the requests that it makes. If it makes multiple requests over separate connections, and the server is able to link those connections somehow, the priority information could be used to manage the precedence of responses across all connections. This thinking also somewhat applies to a protocol translation proxy that talks h2 to the client but HTTP/1.1 to the server. The header is end-to-end (answered on #4) so what does the server do with that header in this case? I think the answer then falls to HTTP core semantics of ignoring unknown headers, in which case we should probably not say anything about a default priority in ignoring case, like you suggest.

draft-kazuho-httpbis-priority.md
LPardue: oops, fixed thanks

draft-kazuho-httpbis-priority.md
kazuho: Regarding what we need to specify in this document, it is my understanding that we only need to talk about endpoints that implement this "extension". We will not be changing the behavior of endpoints that do not implement this, and we should not (unless it is a clarification), because we can't, as we are not updating RFC 723x or 7540.

> I think the answer then falls to HTTP core semantics of ignoring unknown headers, in which case we should probably not say anything about a default priority in ignoring case, like you suggest.

:+1: I think this is the right thing to do.

If we need to make changes to the "Coexistence with HTTP/2 ~" for better clarification can be a separate discussion, but now I realize that it might be better to clarify the rationale a bit more. Below is what I have in mind. Does it make sense to you?

Because the header is end-to-end, it should be used without any negotiation. However, we need to have an exception for HTTP/2, because there is already a different prioritization mechanism _that has wider range of scope_, and we need to provide guidance for transition (or coexistence).

In short term, H2 clients would use both mechanisms, and that would work mostly fine. A server that recognizes both can use either (though we recommend using the header-based scheme), with one exception. The exception is when the client is in fact a proxy that coalesces connections from clients. This is the exact case that we do not cover, and therefore we define a way of ignoring Priority header field in favor of the H2 way.

LPardue: SGTM, I'll update the PR to remove reference to ignoring

LPardue:
Removed use of ignoring header, squashed and updated the PR.

kazuho (approved):

kazuho:
Thank you!

PR #4, opened 2019-06-29
clarify why the header is end-to-end

### LINKED ISSUE #2: Is Priority hop-by-hop?

kazuho:
closes #2.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -212,6 +212,27 @@ the frame-based prioritization scheme defined by HTTP/2.  A server that does not
 recognize this settings parameter would respect the frame-based prioritization
 scheme.
 
+# Considerations
+
+## Why is the Priority header field end-to-end?
+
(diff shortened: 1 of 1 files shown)


LPardue:
This looks great. Now you've written it, it seems almost obvious :)
The thing that was tripping up my thinking was the following, assume a scenario

```
client <-> intermediary <-> server
```
The mistaken way I was thinking occurs at step 2. 

1) the client makes a progressive=yes request. 
2) The intermediary is simple and doesn't process any of the response in chunks, so it makes a request to server with progressive=no. 
3) Once the server completes the response, the intermediary uses the client's progressive=yes directive to manage send prioritization.

The new text clarifies that the Priority header on requests reflects the needs of the true client.

LPardue (approved):

kazuho:
Thank you for checking the PR, as well as the explanation.

I agree that this is confusing, it is good that we are able to clarify the intent in the document.

PR #6, opened 2019-06-30
Change examples to use HTTP/2 style. 

### LINKED ISSUE #5: Modify examples to use H2 syntax?
From a different issue, I realised part of my confusion came from using a HTTP/1.1 example. Perhaps we should consider using a H2 style request in this case, like in rfc 8484

LPardue:
This switches the example style and so fixes #5.

At the same time I spotted a use of a coma ',', which I replaced with a semi-colon ';' to match other parameterised headers.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -61,7 +61,7 @@ file that the document refers to. In contrast, inline images do not block
 rendering and get drawn progressively as the chunks of the images arrive.
 
 To provide meaningful representation of a document at the earliest moment, it is
-important for a HTTP server to prioritize the HTTP responses, or the chunks of
+important for an HTTP server to prioritize the HTTP responses, or the chunks of
 those HTTP responses, that it sends.
 
 HTTP/2 ({{RFC7540}}) provides such a prioritization scheme. A client sends a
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -137,8 +137,11 @@ The following example shows a request for a CSS file with the urgency set to
 `blocking`:
 
 ~~~ example
-GET /style.css HTTP/1.1
-Priority: urgency=blocking
+:method = GET
+:scheme = https
+:authority = example.net
+:path = /style.css
+priority = urgency=blocking
```
kazuho: I think we can remove the trailing empty line (which is part of HTTP/1) now?

draft-kazuho-httpbis-priority.md
```diff
@@ -164,8 +167,11 @@ The following examples shows a request for a JPEG file with the urgency set to
 `non-blocking`, progressive set to `yes`.
 
 ~~~ example
-GET /image.jpg HTTP/1.1
-Priority: urgency=non-blocking, progressive=yes
+:method = GET
+:scheme = https
+:authority = example.net
+:path = /style.css
+priority = urgency=non-blocking, progressive=yes
```
kazuho: Here also.

kazuho (commented):
Thank you for the PR. Looks mostly good to me.

How about adding something like: "The examples of HTTP requests use the HTTP/2-style formatting from {{RFC 7540}}" to Section 1.1?

LPardue:
I obviously made the PR before I had a coffee, so there's some nits I found now. I'll push another update to incorporate your feedback.

LPardue:
PTAL

kazuho:
LGTM! Thanks merged.

Issue #1, opened 2019-06-28
Reprioritization


LPardue:
Reprioritization is an important feature, the document doesn't currently mention it. Since this is a header its likely that what it represents 1) a desired initial priority from the client 2) a server override for its view of the initial priority. Once the response is in flight, the only way to do reprioritization is at the H2 or H3 framing level. 

The doc mentions the SETTINGS_HEADER_BASED_PRIORITY setting that "instructs the server that recognizes the settings parameter to use the header-based prioritization scheme instead of the frame-based prioritization scheme defined by HTTP/2". We might want a different kind of PRIORITY frame to allow reprioritzation in this scenario.

FWIW [priority hints](https://wicg.github.io/priority-hints/) (although a different proposition to this spec) API is based on element attributes and so suffers from lack of reprioritization too. Although that document uses the term, what it means to me is to override the User Agent's view of priority. One potential separate thread of work could consider a more programmatic API that allows management of priority while the response data is in flight.

LPardue:
This also affects the priority of pushed responses, because the client has no means to send a header back

kazuho:
I acknowledge the fact that the draft does not specify a method for a client to change the priority while the response is inflight.

If we are to allow clients reprioritize the requests, I think the answer is obvious. Define a HTTP/2 (or /3) frame that transmits the literal value of the Priority header field. That value would indicate the up-to-date priority of the request. The frame also needs to carry a boolean flag indicating whether the reprioritization request was sent before or after the client saw the response (which could include server's priority response header). Without that information, the server cannot tell if the values found in the response header should override the reprioritization instruction provided by the client.

The question is if we want to define such a frame now. My thought has been that reprioritization is not that important. I think that the only practical case that matters at the moment is for pushed responses, however my view is that servers should prioritize them (asking the browser for providing correct signal loses 1 RTT that we gain from push). The argument against such recommendation would be that people could fail to set correct precedence for push, but it is also true that browsers at the moment do not reprioritize, so meh.

Do you feel that reprioritization is a must-have? If so, I'm positive to defining the frame. If not, I think it might be better to describe the aforementioned design in the Considerations section.

WDYT?

LPardue:
> Do you feel that reprioritization is a must-have?

Personally yes but I think this has come up in various discussion lately (although I could be mis-remembering). 

I think there are uses cases like long lived requests that can benefit, I'm not so bothered by push but it would be a shame to inadvertenly affect it.

[...]

PR #7, opened 2019-07-01
discuss extensibility

kazuho:
I think calling out extension points, at the same time clarifying why it can be left as future work is beneficial for pushing the specification forward.

@LPardue  WDYT?

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -255,14 +255,15 @@ This specification restricts itself to defining just three levels of urgency, in
 order to provide sufficient guranularity for prioritizing responses for ordinary
 web browsing, at minimal complexity.
 
-However, that does not mean that the prioritization scheme would be stuck to
-three levels forvever.  The design provides extensibility.  If deemed necessary,
-it would be possible to split any of the three urgency levels into sub-levels.
+However, that does not mean that the prioritization scheme would forever be
+stuck to the three levels.  The design provides extensibility.  If deemed
+necessary, it would be possible to divide any of the three urgency levels into
+sub-levels.
 
-As an example, a server could assign an `importance` attribute to each image
-that it provides, so that an intermediary could prioritize certain images above
-others.  Or, a graphical user-agent could send a `visible` attribute to indicate
-if the resource being requested is within the viewport.
+As an example, a server could assign an `importance` attribute to the priority
+of each image that it provides, so that an intermediary could prioritize
+certain images above others.  Or, a graphical user-agent could send a `visible`
+attribute to indicate if the resource being requested is within the viewport.
 
 # Security Considerations
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -220,7 +220,7 @@ scheme.
 
 # Considerations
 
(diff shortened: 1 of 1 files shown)


LPardue (approved):
LGTM.

One tangential clarification question. I read the ABNF to allow priority-directive to be either 

```
directive=value
```

or

```
directive
```

Do you think an extension would use the latter syntax? I think your examples would probably use a value.

LPardue:
FWIW another kind of extensibility is a priority-related header. For example, on the Cloudflare blog we explain a `cf-priority` header and a `cf-priority-change header` [1]. 

This allows something like 

```
cf-priority = value #initial value
cf-priority-change = 512:new value # priority value after 512 bytes of response was streamed
```

I'm not suggesting adding something like this now, but it is an example of the flexibility/extensibility of using headers. Furhtermore, the advantage on using headers over frames is that the Web platform has many APIs for Headers. In contrast, HTTP/2 frames have difficulty here.


[1] - https://blog.cloudflare.com/parallel-streaming-of-progressive-images/

kazuho:
Thank you for checking. Merged, as I believe that your comments are better to be solved separately (if necessary).

Regarding the points you have raised, I will respond here in a separate comment.

kazuho:
> One tangential clarification question. I read the ABNF to allow priority-directive to be either
> ```
> directive=value
> ```
> 
> or
> ```
> directive
> ```
> 
> Do you think an extension would use the latter syntax? I think your examples would probably use a value.

I think we should require the values of the parameters to be specified explicitly. IMO that's simpler, and there's no overhead in terms of bytes on the wire once HPACK/QPACK is applied. I think #9 has implies that by stating that the values are sh-token (or sh-boolean), but we can be more explicit.

I will prepare a PR that clarifies that. Please let me know if you prefer allowing omission of values.

kazuho:
Re https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/7#issuecomment-507221127, I totally agree that the `cf-priority-change` header field is a brilliant example of extensibility that a header-based design can provide.

Do you want to mention that in the I-D? The problem in terms of writing it in I-D is that the `cf-priority-change` header is essentially a reprioritization instruction. Therefore, trying to write it down could mean that we'd have certain amount of considerations covering reprioritization (relates to #1).

Regardless of what we decide on I-D, I think we should use `cf-priority-change` as an example on the mailing list / in the presentations.

LPardue:
> I think we should require the values of the parameters to be specified explicitly. IMO that's simpler, and there's no overhead in terms of bytes on the wire once HPACK/QPACK is applied. I think #9 has implies that by stating that the values are sh-token (or sh-boolean), but we can be more explicit.

As an implementer I prefer explicit values.

> Do you want to mention that in the I-D? The problem in terms of writing it in I-D is that the cf-priority-change header is essentially a reprioritization instruction. Therefore, trying to write it down could mean that we'd have certain amount of considerations covering reprioritization (relates to #1).

> Regardless of what we decide on I-D, I think we should use cf-priority-change as an example on the mailing list / in the presentations.

Sorry to have raised that point on this PR. I agree it is related to reprioritization and that we should tackle it as part of #1. I'll add a comment there so we don't lose track. Adding it to the I-D now is a bit distracting regardless of how useful it is to the broader discussion.

PR #8, opened 2019-07-01
Change separator to `,`

kazuho:
I think we need to discuss what the separator is.

ABNF uses `1#priority-directive`, meaning that the separator is `,`. In #6, the examples have been changed to `;` (I thought we were changing examples using `;` to `,`).

My preference is to use `,`, because I think of Priority header to be analogous to Cache-Control header or other header fields using the `1#` rule. I am not sure if there is any reason to use a different rule for the Priority header.

@LPardue WDYT?

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -179,7 +179,7 @@ The following example shows a request for a JPEG file with the urgency set to
 :scheme = https
 :authority = example.net
 :path = /image.jpg
-priority = urgency=non-blocking; progressive=yes
+priority = urgency=non-blocking, progressive=yes
 ~~~
 
 ## Merging Client- and Server-Driven Directives {#merging}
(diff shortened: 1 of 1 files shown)


kazuho:
In relation, we can consider adopting Structured Headers (see #9).

LPardue (commented):
Ahh sorry. I was in an Alt-Svc-style brain mode thinking about parameters. I think comma is better because it could allow composition in the form of separate headers e.g.

```
priority = urgency=non-blocking
priority = progressive=yes
```

This may allow for a simple way to better compression ala mnot (https://www.mnot.net/blog/2018/11/27/header_compression). Not that we should be optimising at this stage but it can help us reason about #9 too.

LPardue (approved):

kazuho:
Thank you for the quick responses. I agree to your points. #9 has subsumed this PR.

PR #9, opened 2019-07-01
use Structured Headers

kazuho:
Use Structured Headers for header field value representation.

The benefit is that it becomes more "extensible" than the current ABNF-based design. The downside could be the introduction of the dependency against an I-D, but hopefully S-H would already be stable for the small portion of the syntax that we use.


LPardue (approved):
+1 for this change. Let's embrace the new best practice.

PR #11, opened 2019-07-03
clarify additions to IANA registries, make SETTINGS_HEADER_BASED_PRIORITY a boolean settings parameter

kazuho:
Writes down required changes to the IANA registries.

While doing that, I noticed that using the existence of a settings parameter to indicate something does not follow the practice in HTTP/2, therefore changed SETTINGS_HEADER_BASED_PRIORITY to a boolean value.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -197,15 +197,19 @@ parameter.
 When connecting to an HTTP/2 ({{!RFC7540}}) server, a client that uses this
 header-based prioritization scheme SHOULD send a
 `SETTINGS_HEADER_BASED_PRIORITY` settings parameter (0xTBD) with a value of
-zero.  An intermediary SHOULD set the settings parameter for a connection it
+one.
+
+An intermediary SHOULD set this settings parameter to one for a connection it
 establishes when and only when all the requests to be sent over that connection
 originates from a client that utilizes this header-based prioritization scheme.
+Otherwise this settings parameter SHOULD be set to zero.
 
-The existence of this settings parameter instructs the server that recognizes
-the settings parameter to use the header-based prioritization scheme instead of
-the frame-based prioritization scheme defined by HTTP/2.  A server that does not
-recognize this settings parameter would respect the frame-based prioritization
-scheme.
+The value of this settings parameter set to a non-zero value instructs the
+server that recognizes the settings parameter to use the header-based
+prioritization scheme instead of the frame-based prioritization scheme defined
+in HTTP/2.  The frame-based prioritization scheme is respected when the value of
+this settings parameter is set to zero, or when the server does not recognize
+the settings parameter.
 
 # Considerations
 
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -260,6 +264,22 @@ TBD
...
+
+* Header field name: Priority
+* Applicable protocol: http
+* Status: standard
+* Author/change controller: IETF
+* Specification document(s): [this document]
+* Related information:
+
+This specification registers the following entry in the HTTP/2 Settings registry
+established by {{?RFC7540}}:
```
LPardue, line 278: informative reference will conflict with normative?

LPardue (commented):
WDYT about a slightly restructured approach to this? I can make a PR if it is easier to understand.

* _Start the section with the status quo / problem statement, roughly_
  * As standard, HTTP/2 endpoints use the frame-based prioritization scheme. Supporting endpoints can substitute this with the header-prioritization scheme. Although it is safe to use header-based prioritization without prior knowledge of a peer's support, it is RECOMMENDED that it be negotiated by means of an HTTP/2 setting.
    * _Negotiate here can be uni-lateral or bi-lateral. Today I think you intend this to be uni-lateral._
* The new settings parameter is SETTINGS_HEADER_BASED_PRIORITY. Its value must be 0 or 1; its initial value is 0.
  * Frame-based prioritization is respected when the value of this setting is 0, or when the server does not recognize the setting.
  * (new bilateral) A server MAY advertise support for header-based priority by sending this setting with a value of 1.
 * A client that uses header-based priority SHOULD send this setting with a value of 1.
 * _Intermediary rules unchanged_ - "An intermerdiary SHOULD set this..."
 * A sender MUST NOT send a SETTINGS_HEADER_BASED_PRIORITY parameter
   with the value of 0 after previously sending a value of 1.
   * _no ability to transition back and forth between the two schemes during a session!_

draft-kazuho-httpbis-priority.md
kazuho: At least my build passes, and RFC7540 appears as normative in the generated I-D. My assumption that normative is given preference when there is a conflict.

kazuho:
@LPardue I very much appreciate a PR, as I believe that you have a particular structure of the section in mind (FWIW what you describe sounds good to me), and because I trust your writing skill more than mine :-)

OTOH, I might have some comments regarding the design changes.

> (new bilateral) A server MAY advertise support for header-based priority by sending this setting with a value of 1.

I am not sure if there is a reason for the server to advertise the setting. Priority header is an end-to-end header. Therefore I think that a client SHOULD send the Priority header regardless of if the server (could be intermediary) it connects to recognizes the header.

Note that we need a server-advertised settings value _if_ we are going to support reprioritization indicated by a frame that uses QPACK, because the client and the server needs to agree on which frames affect the QPACK state. But we need an explicit signal for that purpose. An existing signal cannot be changed at a later moment to indicate that.

Therefore, I think that "a server MUST NOT send SETTINGS_HEADER_BASED_PRIORITY" might be better direction.

> A sender MUST NOT send a SETTINGS_HEADER_BASED_PRIORITY parameter with the value of 0 after previously sending a value of 1.

:+1: We need this.

LPardue:
> I very much appreciate a PR, as I believe that you have a particular structure of the section in mind (FWIW what you describe sounds good to me), and because I trust your writing skill more than mine :-)

Your skills are great. It's a simpler task to iterate on text that is on the page - you make my job easier ;)

I think we should land this PR and then I can open a separate one with my editorial proposal text. And a separate issue to track any proposed design changes.

LPardue (approved):

kazuho:
> I think we should land this PR and then I can open a separate one with my editorial proposal text. And a separate issue to track any proposed design changes.

:+1: Merged as you typed :-)

Issue #13, opened 2019-07-03
Blocking vs non-blocking ordering


pmeenan:
One fairly common pattern is for there to be blocking scripts in the head as well as at the end of the document with imagers in between (with some of the images in the viewport).

Technically both the scripts at the beginning and the end are both blocking but the visible images may be of a higher importance to deliver than the scripts at the end.  If we recommend that all of the blocking resources be delivered before non-blocking it would invert those (granted it is only a SHOULD but implementing to the spec might lead to worse performance in that case).

kazuho:
Thank you for spotting the issue!

I think we have two options.

The first is to replace "processing" with "using". I think the problem is in discussing if "processing" of the document is blocked. If we change that to a discussion of if the "use" of the document is blocked, then I think the text would naturally imply that scripts in body should not be given `urgency=blocking`.

The other option is to choose a different term. One attractive choice is to reuse the terms defined in Priority Hints, but the question is if they would match nicely to the model we need in HTTP prioritization. As we have been discussing in #12, importance=high does not maps (or should not be mapped) to urgency=blocking.

kazuho:
Replying to myself:
> The first is to replace "processing" with "using".

#14 does that. I think it would be a better approach than changing the terms to "high" / "low", if we are to add things like "deferred" (to indicate that a resource is a script with async attribute).

kazuho:
@pmeenan #14 has been merged. Please close the issue if you are happy with it, otherwise let's discuss other options.

LPardue:
We merged a change to the issue at the time, and subsequent changes to the draft
have rendered this stale. Closing the issue but feel free to open a new one
against the current document.

PR #14, opened 2019-07-04
talk about a response blocking "using a document", rather than "processing"

kazuho:
IMO this would be the minimal fix to #13.

@pmeenan @LPardue Thoughts?

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -110,7 +110,7 @@ Unknown parameters MUST be ignored.
 ## urgency
 
 The `urgency` parameter takes one of the following sh-tokens as the value, that
-indicates how an HTTP response affects the usability of other responses:
+indicates how an HTTP response affects the usage of other responses:
 
 * `blocking` indicates that the response prevents other responses from being
   used.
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -110,14 +110,15 @@ Unknown parameters MUST be ignored.
 ## urgency
 
 The `urgency` parameter takes one of the following sh-tokens as the value, that
-indicates how an HTTP response affects the processing of other responses:
+indicates how an HTTP response affects the usability of other responses:
```
```suggestion
indicates how an HTTP response affects the usage of other responses:
```

usability seems like the wrong word here (it is human-focused). For example, a blocking CSS resource doesn't make the final rendered document less usable (unless it is really bad CSS ;) )

LPardue (changes requested):

draft-kazuho-httpbis-priority.md
kazuho: Thanks you for the suggestion. While I think that we want to be human-focused for the browser use-case, I agree that it might be too nuanced.

LPardue (approved):

LPardue:
oh oops, merged before @pmeenan feedback, sorry!

PR #15, opened 2019-07-04
describe how to serve non-progressive responses

kazuho:
I think that the lack of the text was pointed out by @pmeenan.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -149,8 +149,8 @@ responses that share the same urgency.
 
 A server SHOULD transmit non-progressive responses one by one, preferably in the
 order the requests were generated.  Doing so maximizes the chance of the client
-making progress in processing the composition of the HTTP responses at the
-earliest moments.
+making progress in using the composition of the HTTP responses at the earliest
+moment.
 
 The following example shows a request for a JPEG file with the urgency parameter
 set to `non-blocking` and the progressive parameter set to `1`.
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -121,9 +121,8 @@ indicates how an HTTP response affects the processing of other responses:
 
 The default value is `document`.
 
-A server that supports the Priority header field SHOULD transmit HTTP responses
-in the order of their urgency: `blocking` first, followed by `document`,
-followed by `non-blocking`.
+A server SHOULD transmit HTTP responses in the order of their urgency:
+`blocking` first, followed by `document`, followed by `non-blocking`.
 
 The following example shows a request for a CSS file with the urgency set to
 `blocking`:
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -144,8 +143,13 @@ as chunks of the response arrive.
...
 The default value of the `progressive` parameter is `0`.
 
-A server that supports the Priority header field SHOULD distribute the bandwidth
-of a connection between progressive responses that share the same urgency.
+A server SHOULD distribute the bandwidth of a connection between progressive
+responses that share the same urgency.
+
+A server SHOULD transmit non-progressive responses one by one, in the order the
+requests were received, or if inferable, in the order the requests were
+generated.  Doing so maximizes the chance of the client making progress in
```
LPardue: Reading between the lines I think you are suggesting something around server handling of concurrency.

a) an HTTP/2 server could receive multiple requests at roughly the same time instant. However, it can infer the request precedence by looking at the the stream ID; lower stream IDs have higher precedence.

b) an HTTP/3 server could receive requests made at roughly the same instant, in a different order due to independent multiplexing. Again, lower stream IDs have higher precendence.

Is this a correct interpretation. Is it worth being more explicit about this somewhere in the text? (maybe a separate tracking issue).

LPardue: This also makes me wonder if behaviour of intermediaries needs to be described. E.g an intermediary should make upstream requests in the order of generation (if inferable). I'm less convinced this needs to be done TBH.

draft-kazuho-httpbis-priority.md
```diff
@@ -144,8 +143,13 @@ as chunks of the response arrive.
...
+processing the composition of the HTTP responses at the earliest moments.
```
LPardue: due to #14, maybe you want to say something different to "processing the composition"

LPardue (changes requested):

draft-kazuho-httpbis-priority.md
kazuho: > Is this a correct interpretation.

Yes.

> Is it worth being more explicit about this somewhere in the text?

Nice catch! By reading your comments, I think I might have written too much.

As you say, in case of HTTP/2, the order in which the client issued the requests is preserved in the order the server receives them. So discussing about using the receiving order and also suggesting the use of stream IDs is as detailed as having recommendations specific to HTTP/2 and HTTP/3. Considering that, I think it would be natural to wonder why we can't name the protocols and discuss more precisely.

To paraphrase, I think that the text was too specific even though it did not name the protocols. Therefore, I've trimmed down the suggestion in 92bef73.

That said, if you prefer, I think we can discuss how the guideline (i.e. send in the order the requests were generated) can be implemented in HTTP/2 and HTTP/3. Though the downside would be, that there'd be the next question as you point out: what about proxies?

LPardue: yes lets separate this out! I think it is worth having the text because it relates to some of the issues around FIFO processing that @rmarx has presented.

draft-kazuho-httpbis-priority.md
kazuho: Good point.

So, should we merge #14? I think #14 to be an improvement. Then, we can change "processing" that appears here to "using" and merge this PR as well?

kazuho: As I write here, I notice #14 getting merged. Thank you! Making edits to this PR.

draft-kazuho-httpbis-priority.md
kazuho: :+1:

LPardue (approved):

kazuho:
Thank you for checking!

PR #16, opened 2019-07-04
replace bullets with list syntax and fix link lint

LPardue:
Use list style. fixes `- No link definition for link ID 'this document' found on line 325` message.


kazuho (approved):
LGTM. Avoiding compiler warnings is preferable.

I had accustomed myself to using `[this specification]` but looking at some of the httpbis drafts that became RFCs, it seems that some do that, while some don't.

PR #18, opened 2019-07-04
restructure the text related to H2 coexistence. Make SETTINGS usage …

LPardue:
…clearer.

To start, we don't use the terms header-based and frame-based until late in the document. So I pulled that straight up into the introduction.

The H2 coexistence section then builds on that earlier definition. Promoted definition of the SETTING to a subsection for more visibility and got stricter on the rules for using it.

Bottom-line: the parameter is a recommended useful signal for a client to use when it wants to use header-based prioritisation. However, no negotiation is required.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -77,7 +77,9 @@ bandwidth among the HTTP responses.  However, the design has shortcomings:
 Based on these observations, this document defines the Priority HTTP header
 field that can be used by both the client and the server to specify the
 precedence of HTTP responses in a standardized, extensible, protocol-version-
-independent, end-to-end format.
+independent, end-to-end format. This header-based prioritization scheme can act
+as a substitute for the HTTP/2 frame-based prioritization scheme (see
+{{coexistence}}).
 
 ## Notational Conventions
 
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -197,24 +199,38 @@ because the server-provided value overrides that provided by the client.  The
...
-# Coexistence with HTTP/2 Priorities
+# Coexistence with HTTP/2 Priorities {#coexistence}
 
-When connecting to an HTTP/2 ({{!RFC7540}}) server, a client that uses this
-header-based prioritization scheme SHOULD send a
-`SETTINGS_HEADER_BASED_PRIORITY` settings parameter (0xTBD) with a value of
-one.
+Standard HTTP/2 ({{!RFC7540}}) endpoints use frame-based prioritization, whereby
+a client sends priority information in dedicated fields present in HEADERS and
+PRIORITY frames. A client may instead choose to use header-based prioritization
```
kazuho: may -> might or can?

kazuho (approved):
This looks much better. Thank you for writing it down.

LGTM modulo the nit-pick below.

draft-kazuho-httpbis-priority.md
LPardue: Gah I keep doing this, want the lint tool to check it!

Will fix up soon.

PR #19, opened 2019-07-05
servers can use signals other than the Priority header field

kazuho:
I think it's a good idea to clarify that a server can use other signals to supplement the information carried by the Priority header field.

I thought that we'd better point that out, at least in the considerations where we describe possibilities of extending the Priority field.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -263,6 +263,12 @@ of each image that it provides, so that an intermediary could prioritize
 certain images above others.  Or, a graphical user-agent could send a `visible`
 attribute to indicate if the resource being requested is within the viewport.
 
+Also, to improve the prioritization logic, a server can use signals other than
+that carried by the Priority header field. For example, a server might
+prioritize a response of a font {{?RFC8081}} above images that share the same
+urgency, so that a visual client can render textual information at an early
+moment.
+
 # Security Considerations
 
 TBD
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -263,6 +263,12 @@ of each image that it provides, so that an intermediary could prioritize
 certain images above others.  Or, a graphical user-agent could send a `visible`
 attribute to indicate if the resource being requested is within the viewport.
 
+Also, to improve the prioritization logic, a server can use signals other than
+that carried by the Priority header field. For example, a server might
```
LPardue: I want to check that my reading of this paragraph is what was intended. The preeceeding paragraphs talks about "in-band" signals in the form of Priority paramaters. Is this paragraph describing "out-of-band" signals? Do you think this rewording might be clearer?

_A server can combine the hints provided in the Priority header field with other information in order to improve the prioritization of responses. For example, a server that receives requests for a font {{?RFC8081}} and images with the same urgency might allocated more resources to the font, so that a visual client can render textual information at an early moment._

kazuho: That's what I meant and your sentence is better.

I've applied it with one change: I've replaced "allocate more resources" to "give higher precedence", as the former might imply that the bandwidth is shared between the responses rather than strictly ordered. I think we should allow both weight-based and dependency-based optimization within the same urgency.

LPardue (approved):
LGTM

PR #22, opened 2019-07-05
state SH terms in notational conventions

LPardue:
I'm still getting familiar with Structured headers, so lets state where the terms sh-token and sh-boolean come from.


kazuho:
Makes sense.

PR #25, opened 2019-07-08
Change initial to default

rmarx:
Seems like that was more the intent

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -212,7 +212,7 @@ prioritization as specified in this document.
 To improve communication of the client's intended prioritization scheme, this
 document specifies a new HTTP/2 SETTINGS parameter with the name
 `SETTINGS_HEADER_BASED_PRIORITY`. The value of the parameter MUST be 0 or 1; the
-initial value is 0. Frame-based prioritization is respected when the value is 0,
+default value is 0. Frame-based prioritization is respected when the value is 0,
 or when the server does not recognize the setting.
 
 An HTTP/2 client that uses header-based priority SHOULD send a


LPardue:
The term `initial` was used on purpose to align with HTTP/2 section 5.5 e.g

```
   This document doesn't mandate a specific method for negotiating the
   use of an extension but notes that a setting (Section 6.5.2) could be
   used for that purpose.  If both peers set a value that indicates
   willingness to use the extension, then the extension can be used.  If a 
   setting is used for extension negotiation, the initial value MUST
   be defined in such a fashion that the extension is initially
   disabled.
```

rmarx:
I am all for consistency :) Just feels weird, as IIUC in H3 there is no way to update this setting once the connection is established. Or maybe this is meant more to indicate connection resumption?

LPardue:
Well now I see that H2 and H3 are inconsistent :(

From H2
```
SETTINGS_INITIAL_WINDOW_SIZE (0x4):  Indicates the sender's initial
      window size (in octets) for stream-level flow control.  The
      initial value is 2^16-1 (65,535) octets.
```

From H3
```
SETTINGS_MAX_HEADER_LIST_SIZE (0x6):  The default value is unlimited.
      See Section 5.1.1 for usage.
```
also
```
    All settings begin at an initial value, and are updated upon receipt
   of a SETTINGS frame.  For servers, the initial value of each client
   setting is the default value.
```

So, now I'm none the clearer on which is the better term.

kazuho:
I prefer “Initial”.

An H2/H3 connection becomes usable before receiving a SETTINGS frame from the peer. Until then, the initial value is applied. It doesn’t matter for this particular settings value, though.

LPardue:
The authors preferred initial, and this PR is rotted. Closing but reopen if you fee a strong need to try and convince us .

PR #29, opened 2019-07-10
Expand the description of urgencies

kazuho:
Aside from the use of HTTP headers, the semantic urgency levels is the core of the proposal, but we've been very brief about what each level represents.

This PR expands on that.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -150,10 +150,10 @@ precedence as the contents of the document.
 
 For example, when a user using a web browser navigates to a different HTML
 document, the request for that HTML is given the `document` urgency.  When that
-HTML document uses a custom font, the HTTP request for that custom font SHOULD
-also be given the `document` urgency.  This is because the availablity of the
-custom font is the precondition for the user to use the portion of the HTML
-document, which is to be rendered by that font.
+HTML document uses a custom font, the request for that custom font SHOULD also
+be given the `document` urgency.  This is because the availablity of the custom
+font is the precondition for the user to use the portion of the HTML document,
+which is to be rendered by that font.
 
 ### non-blocking
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -150,7 +150,7 @@ precedence as the contents of the document.
 
 For example, when a user using a web browser navigates to a new HTML document,
 the request for that HTML is given the `document` urgency.  When that HTML
-document uses a custom font, HTTP request for that custom font SHOULD also be
+document uses a custom font, the HTTP request for that custom font SHOULD also be
 given the `document` urgency, because it is generally assumed that the usability
 of the HTML document partially depends on the availability of the font, even
 though the font does not block the use of the document as a whole.
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -151,9 +151,9 @@ precedence as the contents of the document.
 For example, when a user using a web browser navigates to a new HTML document,
 the request for that HTML is given the `document` urgency.  When that HTML
 document uses a custom font, HTTP request for that custom font SHOULD also be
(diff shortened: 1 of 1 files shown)


rmarx:
I think these clarifications help a lot in explaining the intended usage and semantics of the current urgency values

draft-kazuho-httpbis-priority.md
```diff
@@ -139,6 +131,39 @@ The following example shows a request for a CSS file with the urgency set to
...
+
+### document
+
+The `document` urgency indicates that the response contains the document that is
+being processed.  This urgency is also used for responses that deserve the same
+precedence as the contents of the document.
+
+For example, when a user using a web browser navigates to a new HTML document,
+the request for that HTML is given the `document` urgency.  When that HTML
+document uses a custom font, HTTP request for that custom font SHOULD also be
```
```suggestion
document uses a custom font, the HTTP request for that custom font SHOULD also be
```

draft-kazuho-httpbis-priority.md
```diff
@@ -139,6 +131,39 @@ The following example shows a request for a CSS file with the urgency set to
...
+given the `document` urgency, because it is generally assumed that the usability
+of the HTML document partially depends on the availability of the font, even
```
LPardue: I'm not sure how to realize what "partially depends" means.

draft-kazuho-httpbis-priority.md
```diff
@@ -139,6 +131,39 @@ The following example shows a request for a CSS file with the urgency set to
...
+though the font does not block the use of the document as a whole.
+
+### non-blocking
+
+The `non-blocking` urgency indicates that the response does not prevent the
+client from using the document even though the response is being used or
+referred to by the document.
```
LPardue: I think this is accurate but its easy to misread and overlook the difference between "client using document" and "document using resource". I don't have a suggested alternate wording though.

draft-kazuho-httpbis-priority.md
```diff
@@ -139,6 +131,39 @@ The following example shows a request for a CSS file with the urgency set to
...
+
+For example, inline images typically improves the usability of an HTML document.
```
LPardue: either:

a) `inlining images typically improves`
b) `inline image typically improve`

LPardue (changes requested):

LPardue:
I think these changes are a net positive. 

However, these shine light on what I think are some assumptions about the reader's understanding of usability, especially in relation to User Agent's parse/render workflows and the way they might cope with fallbacks.

I think some text in a separate PR could help describe usability. Which might help contrast this prioritisation model against frame- / tree-based. What I mean by this is that H2's model is not subtle enough to articulate a priority of "I'd really like this thing, but if I really had to pick, I could probably manage without it." - or maybe it could but it would incur chatiness and reprioritization.

draft-kazuho-httpbis-priority.md
kazuho: I've tried to explain that in 95f1766. I'm not sure if I've succeeded though.

draft-kazuho-httpbis-priority.md
kazuho: When for "incorporate"in 5c319d7. Please let me know what you think.

draft-kazuho-httpbis-priority.md
kazuho: Went for plural in d76ba24.

kazuho:
@LPardue Thank you for your comments.

> However, these shine light on what I think are some assumptions about the reader's understanding of usability, especially in relation to User Agent's parse/render workflows and the way they might cope with fallbacks.

Agreed. While writing down this PR and #30, I intentionally tried to imply that there is some wiggle room in how certain resources (e.g., font) might be generally prioritized by a client, and then suggest how they can be overridden by the server.

But yes, we should try to make it better.

But as you state, making such improvements might take time, and therefore I share the view that we'd better merge PR (or the sort) then do further work on a different PR. In short, please give the updated text a review.

kazuho:
This PR was subsumed by #34.

PR #30, opened 2019-07-10
explain server overrides using urgency

kazuho:
Current text explains server override by using the progressive parameter as the example.

However, it seems to be confusing (see https://lists.w3.org/Archives/Public/ietf-http-wg/2019JulSep/0042.html).

It would be better to talk about changing urgency rather than progressiveness, because the former would be the common use-case and has bigger impact on user experience.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -171,7 +171,7 @@ priority = urgency=non-blocking, progressive=?1
 It is not always the case that the client has the best understanding of how the
 HTTP responses deserve to be prioritized.  For example, use of an HTML document
 might depend heavily on one of the inline images.  Existence of such dependency
-is typcially best known by the person that administers the document.
+is typically best known to the person that administers the document.
 
 By using the "Priority" response header, a server can override the
 prioritization hints provided by the client.  When used, the parameters found
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -168,14 +168,14 @@ priority = urgency=non-blocking, progressive=?1
 
 # Merging Client- and Server-Driven Parameters {#merging}
 
-It is not always the case that the client has the best view of how the HTTP
-responses should be prioritized.  For example, whether a JPEG image should be
-served progressively by the server depends on the structure of that image file
-- a property only known to the server.
+It is not always the case that the client has the best understanding of how the
+HTTP responses deserve to be prioritized.  For example, use of an HTML document
+might depend heavily on one of the inline images.  Existence of such dependency
+is typcially best known by the person that administers the document.
 
-Therefore, a server is permitted to send a "Priority" response header field.
-When used, the parameters found in this response header field override those
-specified by the client.
+By using the "Priority" response header, a server can override the
+prioritization hints provided by the client.  When used, the parameters found
+in the response header field overrides those specified by the client.
 
 For example, when the client sends an HTTP request with
 
@@ -183,7 +183,7 @@ For example, when the client sends an HTTP request with
 :method = GET
 :scheme = https
 :authority = example.net
-:path = /image.jpg
+:path = /menu.png
 priority = urgency=non-blocking, progressive=?1
 ~~~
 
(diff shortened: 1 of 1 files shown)


kazuho:
@rmarx I do not assume that you'd agree with the approach, but I think this PR and #29 would better clarify the intended use of urgencies.

rmarx:
You are correct, I do not agree with this approach :) 

The example you list here is exactly the type of thing that I think will lead to much confusion for end users... Pretending an image is a document, akin to HTML, to coerce server-side logic to give it a higher priority, feels like you're "hacking" an existing system to exhibit new behaviour, which is weird for a fully new spec. 

That being said, if you do want to go this route, this change makes things a lot clearer about the intent and potential ways of use, so it's a good PR from that perspective imo.

draft-kazuho-httpbis-priority.md
```diff
@@ -168,37 +168,37 @@ priority = urgency=non-blocking, progressive=?1
 
 # Merging Client- and Server-Driven Parameters {#merging}
 
-It is not always the case that the client has the best view of how the HTTP
-responses should be prioritized.  For example, whether a JPEG image should be
-served progressively by the server depends on the structure of that image file
-- a property only known to the server.
+It is not always the case that the client has the best understanding of how the
+HTTP responses deserve to be prioritized.  For example, use of an HTML document
+might depend heavily on one of the inline images.  Existence of such dependency
```
```suggestion
might depend heavily on one of the inline images.  Existence of such dependencies
```

draft-kazuho-httpbis-priority.md
```diff
@@ -168,37 +168,37 @@ priority = urgency=non-blocking, progressive=?1
...
+is typically best known to the person that administers the document.
```
LPardue: "person that administers" sounds like an odd turn of phrase to me. I think you were shooting for replacing implementation-oriented "server" but I don't know if this hits the mark. I have no other suggestions though!

draft-kazuho-httpbis-priority.md
```diff
@@ -168,37 +168,37 @@ priority = urgency=non-blocking, progressive=?1
...
 ~~~
 
-the intermediary's view of the progressiveness of the response becomes negative,
-because the server-provided value overrides that provided by the client.  The
-urgency is deemed as `non-blocking`, because the server did not specify the
-parameter.
+the intermediary's view of the urgency of the response is promoted from
+`non-blocking` to `document`, because the server-provided value overrides that
+provided by the client.  The progressiveness of the response continues to be
+deemed as true, because the server did not specify the `progressive` parameter.
```
LPardue: This confuses me as it seems to conflict with previous guidance of

```
This document defines the urgency and progressive parameters. Values of these parameters MUST always be present. When any of the defined parameters are omitted, or if the Priority header field is not used, their default values SHOULD be applied.
```

and

```
The default value of the progressive parameter is 0.
```

Are you suggesting a third case where omission in responses means the request value is respected?

draft-kazuho-httpbis-priority.md
kazuho: I agree that "server" is better. Fixed in 21624c7.

draft-kazuho-httpbis-priority.md
kazuho: > Are you suggesting a third case where omission in responses means the request value is respected?

Yes. I think I've better clarified that in 9f30235.

kazuho:
Thank you for the suggestions. Please give it another try.

draft-kazuho-httpbis-priority.md
LPardue: I think the new text clarifies this example but I'm not sure that a reader of the parameter definition spec would come to the conclusion that this is the expected behaviour. In my mind, the rules for priority header were quite symmetric but I think this highlights an asymmetry between how they are applied across requests and responses. I think we might want to spin that out into a separate issue, WDYT?

LPardue (approved):
LGTM modulo the other comment I made (which we should not try to address on this PR).

draft-kazuho-httpbis-priority.md
kazuho: Opening a separate issue makes sense to me.

kazuho:
Thank you for the review, and for letting is pass through keeping the debatable point separately.

PR #34, opened 2019-07-11
8 urgency levels using numbers

kazuho:
In summary, by using the following mapping.

|Urgency|Description|
|---:|:---|
|-1|blocking|
|0|document|
|1 to 5|non-blocking|
|6|background|

Within the non-blocking levels, 2, 3, 4 are expected to be used by the client, where 3 is the default, 2 is high, 4 is low. 1 and 5 are expected to be used by the server for intervention.

We could also consider rephrasing "document" to "default."

Builds on top of PR #29.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -150,10 +150,10 @@ precedence as the contents of the document.
 
 For example, when a user using a web browser navigates to a different HTML
 document, the request for that HTML is given the `document` urgency.  When that
-HTML document uses a custom font, the HTTP request for that custom font SHOULD
-also be given the `document` urgency.  This is because the availablity of the
-custom font is the precondition for the user to use the portion of the HTML
-document, which is to be rendered by that font.
+HTML document uses a custom font, the request for that custom font SHOULD also
+be given the `document` urgency.  This is because the availablity of the custom
+font is the precondition for the user to use the portion of the HTML document,
+which is to be rendered by that font.
 
 ### non-blocking
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -173,7 +173,13 @@ The non-blocking urgency indicates that the response does not prevent the client
 from using the document even though the response is being incorporated to or
 referred to by the document.
 
-For example, inline images typically improve the usability of an HTML document.
+For example, inline images (i.e., images being fetched and displayed as part of
+the document, typically through the use of the IMG tag) are visually important
+elements of an HTML document, however their unavailability is not expected to
+prevent users from using using the document.  Display of those images is
+considered to be an improvement for visual clients rather than a prerequisite
+for all user agents.
+
 Therefore, such images will be given the non-blocking urgency.
 
 Values between 1 and 5 are used to represent this urgency, to provide wiggle
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -150,7 +150,7 @@ precedence as the contents of the document.
 
 For example, when a user using a web browser navigates to a new HTML document,
 the request for that HTML is given the `document` urgency.  When that HTML
-document uses a custom font, HTTP request for that custom font SHOULD also be
+document uses a custom font, the HTTP request for that custom font SHOULD also be
 given the `document` urgency, because it is generally assumed that the usability
 of the HTML document partially depends on the availability of the font, even
 though the font does not block the use of the document as a whole.
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -256,9 +256,9 @@ priority = urgency=1
 ~~~
 
 the intermediary's understanding of the urgency is promoted from `3` to `1`,
-because the server-provided value overrides that provided by the client.  The
-progressiveness continues to be `1`, the value specified by the client, as the
-server did not specify the `progressive` parameter.
+because the server-provided value overrides the value provided by the client.
+The progressiveness continues to be `1`, the value specified by the client, as
+the server did not specify the `progressive` parameter.
 
 # Coexistence with HTTP/2 Priorities {#coexistence}
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -192,6 +192,11 @@ prioritize a response above or below other non-blocking responses.
 Clients MAY use values 2 and 4 to indicate that a request is given relatively
 higher or lower priority within the non-blocking urgency group.
 
+For example, an image certain to be visible at the top of the page, might be
+assigned a value of 2 instead of 3, as it will have a high visual impact for the
+user.  Conversely, an asynchronously loaded JavaScript file might be assigned an
+urgency value of 4, as it will have no visual impact.
+
 When none of the considerations above is applicable, the value of 3 SHOULD be
 used.
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -151,9 +151,9 @@ precedence as the contents of the document.
 For example, when a user using a web browser navigates to a new HTML document,
 the request for that HTML is given the `document` urgency.  When that HTML
 document uses a custom font, HTTP request for that custom font SHOULD also be
-given the `document` urgency, as the usability of the HTML document partially
-depends on the availability of the font, even though the font does not block
-the use of the document as a whole.
+given the `document` urgency, because it is generally assumed that the usability
+of the HTML document partially depends on the availability of the font, even
+though the font does not block the use of the document as a whole.
 
 ### non-blocking
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -148,9 +148,8 @@ being used.
 
 Value of -1 is used to represent this urgency.
 
-For example, use of an external stylesheet blocks a web browser from rendering
-the HTML, when the stylesheet is being referred to from the HEAD section of the
-HTML.  In such case, the stylesheet is given the blocking urgency.
+For example, use of an external stylesheet can block a web browser from
+rendering the HTML. In such case, the stylesheet is given the blocking urgency.
 
 ### document
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -143,21 +143,17 @@ Endpoints SHOULD respect the definition of the values when assigning urgencies.
 
 ### blocking
 
-The blocking urgency indicates that the response prevents other responses from
-being used.
-
-Value of -1 is used to represent this urgency.
+The blocking urgency (value -1) indicates that the response prevents other
+responses from being used.
 
 For example, use of an external stylesheet can block a web browser from
 rendering the HTML. In such case, the stylesheet is given the blocking urgency.
 
 ### document
 
-The document urgency indicates that the response contains the document that is
-being processed.  This urgency is also used for responses that deserve the same
-precedence as the contents of the document.
-
-Value of zero (0) is used to represent this urgency.
+The document urgency (value 0) indicates that the response contains the document
+that is being processed.  This urgency is also used for responses that deserve
+the same precedence as the contents of the document.
 
 For example, when a user using a web browser navigates to a different HTML
 document, the request for that HTML is given the document urgency.  When that
@@ -201,10 +197,8 @@ used.
 
 ### background
 
-The background urgency is used for responses of which the delivery can be
-postponed without having an impact on using other responses.
-
-Value of 6 is used to represent this urgency.
+The background urgency (value 6) is used for responses of which the delivery can
+be postponed without having an impact on using other responses.
 
 As an example, a software update running in the background would be assigned the
 background urgency.
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -148,7 +148,7 @@ being used.
 
 Value of -1 is used to represent this urgency.
 
-For example, a stylesheet being referred from the HEAD section of an HTML
+For example, a stylesheet being referred to from the HEAD section of an HTML
 document blocks a web browser from rendering the HTML.  In such case, the
 stylesheet is given the blocking urgency.
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -158,8 +158,8 @@ document, which is to be rendered by that font.
 ### non-blocking
 
 The `non-blocking` urgency indicates that the response does not prevent the
-client from using the document even though the response is being used or
-referred to by the document.
+client from using the document even though the response is being incorporated to
+or referred to by the document.
 
 For example, inline images typically improve the usability of an HTML document.
 Therefore, such images will be given the `non-blocking` urgency.
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -112,18 +112,10 @@ Unknown parameters MUST be ignored.
 
 ## urgency
 
-The `urgency` parameter takes one of the following sh-tokens as the value that
-indicates how an HTTP response affects the usage of other responses:
+The `urgency` parameter takes one of the following sh-tokens as the value:
+`blocking`, `document`, `non-blocking`.  The default value is `document`.
 
-* `blocking` indicates that the response prevents other responses from being
-  used.
-* `document` indicates that the response contains the document that is being
-  processed.
-* `non-blocking` indicates that the response does not prevent the client from
-  using the document even though the response is being used or referred to by
-  the document.
-
-The default value is `document`.
+The values indicate how an HTTP response affects the usage of other responses.
 
 A server SHOULD transmit HTTP responses in the order of their urgency:
 `blocking` first, followed by `document`, followed by `non-blocking`.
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -184,21 +244,21 @@ For example, when the client sends an HTTP request with
...
-priority = urgency=document
+priority = urgency=1
 ~~~
 
-the intermediary's understanding of the urgency is promoted from `non-blocking`
-to `document`, because the server-provided value overrides that provided by the
-client.  The progressiveness continues to be `1`, the value specified by the
-client, as the server did not specify the `progressive` parameter.
+the intermediary's understanding of the urgency is promoted from `3` to `1`,
+because the server-provided value overrides that provided by the client.  The
```
rmarx: > overrides that

overrides the one

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,93 @@ Unknown parameters MUST be ignored.
...
+Endpoints SHOULD respect the definition of the values when assigning urgencies.
+
+### blocking
+
+The blocking urgency indicates that the response prevents other responses from
+being used.
+
+Value of -1 is used to represent this urgency.
+
+For example, a stylesheet being referred from the HEAD section of an HTML
```
rmarx: > being referred from

being referred to from
being referred to by

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,93 @@ Unknown parameters MUST be ignored.
...
+document blocks a web browser from rendering the HTML.  In such case, the
```
rmarx: > from rendering the HTML

from rendering the remainder of the HTML

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,93 @@ Unknown parameters MUST be ignored.
...
+being processed.  This urgency is also used for responses that deserve the same
+precedence as the contents of the document.
+
+Value of zero (0) is used to represent this urgency.
+
+For example, when a user using a web browser navigates to a different HTML
+document, the request for that HTML is given the document urgency.  When that
+HTML document uses a custom font, the request for that custom font SHOULD also
+be given the document urgency.  This is because the availablity of the custom
+font is the precondition for the user to use the portion of the HTML document,
```
rmarx: > is the precondition

is likely a precondition
(because nowadays you have things like font-display, which allows developers to specify a fallback)

> to use the portion of the HTML

to use that portion of the HTML

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,93 @@ Unknown parameters MUST be ignored.
...
+which is to be rendered by that font.
+
+### non-blocking
+
+The non-blocking urgency indicates that the response does not prevent the client
+from using the document even though the response is being incorporated to or
+referred to by the document.
+
+For example, inline images typically improve the usability of an HTML document.
+Therefore, such images will be given the non-blocking urgency.
```
rmarx: As remarked by Lucas before, this passage feels weird... what exactly are "inline" images here? And how does them improving the usability of a document lead to them being non-blocking?

Suggestion:
For example, images referred to in an HTML document can typically be processed and rendered independently from other resources. Therefore, such images will be given the non-blocking urgency.

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,93 @@ Unknown parameters MUST be ignored.
...
+
+Values between 1 and 5 are used to represent this urgency, to provide wiggle
+room for endpoints to give some responses more or less precedence than others
+that belong to the non-blocking group.
+
+Clients SHOULD NOT use values 1 and 5.  Servers MAY use these values to
+prioritize a response above or below other non-blocking responses.
+
+Clients MAY use values 2 and 4 to indicate that a request is given relatively
+higher or lower priority within the non-blocking urgency group.
```
rmarx: if you agree with the comment about the "inline images" above, then here you might add something like:

For example, an image certain to be visible at the top of the page, might be assigned a value of 4 instead of 3, as it will have a high visual impact for the user. 

And if you're feeling verbose:
Conversely, smaller images such as icons or images lower on the page that have less immediate visual impact can be assigned an urgency of value 2. 

To complete the cycle of image-based examples:
Finally, if there is a single, large and important image known to be on the top of the page (sometimes called a hero image), the server can assign it a value of 5 to make sure it is shown to the user as soon as possible.

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,93 @@ Unknown parameters MUST be ignored.
...
+
+### background
+
+The background urgency is used for responses of which the delivery can be
+postponed without having an impact on using other responses.
+
+Value of 6 is used to represent this urgency.
+
+As an example, a software update running in the background would be assigned the
+background urgency.
```
rmarx, line 205: I don't really like this example for the web browsing use case, which you reference later in the document (see the "ordinary web browsing" part). 

Alternative:
As an example, clients can prefetch resources that are expected to be needed for the next page load. As these will not affect the the current page, they would all be assigned the background urgency.

Better with the final addition:
. If these prefetched resources are still in flight when the navigation to the next page happens, they can be re-prioritized to their actual urgency values.

draft-kazuho-httpbis-priority.md
kazuho: Would "one" be confusing with "`1`" being used right before?

Maybe it would be best to simply write "because the server-provided value overrides the client-provided value".

draft-kazuho-httpbis-priority.md
kazuho: Doesn't the CSS, by default, block the first paint?

I could well be wrong, but to me it seems that https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-blocking-css states as such.

draft-kazuho-httpbis-priority.md
kazuho: I'm fine with explaining what "inline images" mean here.

OTOH, I think we need to explain why a custom font, which typically blocks the rendering of some text, should be given higher precedence than an inline image.

It has been my assumption that some images being unavailable is considered to have less impact than some text being unavailable, and that is because an HTML document is considered to be somewhat usable even though images are yet to be rendered. That the rationale for using "improve" here.

I could well be wrong, but we need to discuss the rationale, if we want to suggest that fonts should be given higher precedence than images.

draft-kazuho-httpbis-priority.md
rmarx: ah, I can see how that's confusing :) 

I meant "the one" as referring to the value as a concept, not the value 1. 
As in: the server-provided value overrides the value provided by the client, and not: the server-provided value overrides the 1 provided by the client. 

So maybe just make it "overrides the value" then

draft-kazuho-httpbis-priority.md
rmarx: Yes, I was conflating some things here, good point. CSS in the HEAD indeed blocks the rendering. JS in the HEAD also blocks parsing of the remainder of the HTML, which was probably what I was getting at here. 

However, CSS in the BODY blocks the rendering of the remainder (see https://jakearchibald.com/2016/link-in-body/). 

If the example indeed just mentions CSS in the HEAD, then the original text was correct.

draft-kazuho-httpbis-priority.md
rmarx: I would say your assumptions are correct and that fonts need higher precedence than images.
I just didn't feel the text made that very clear with the "improve the usability" part. 

Besides that, I also feel that people reading this will probably understand that fonts can block text rendering and that text is typically more important than images without explicitly stating it.

draft-kazuho-httpbis-priority.md
kazuho: Makes sense. Thank you for the suggestion. Applied in d7135215.

draft-kazuho-httpbis-priority.md
kazuho: Thank you. Fixed in 4ac3b92.

draft-kazuho-httpbis-priority.md
kazuho: Fixed in 4e0e79e.

draft-kazuho-httpbis-priority.md
kazuho: Thanks. Does ebdf06a make things clearer? Please take a look.

draft-kazuho-httpbis-priority.md
kazuho: Thank you for your thoughts. I've expanded why images are improvements. PTAL at 68f9660.

draft-kazuho-httpbis-priority.md
kazuho: Thank you for the suggestions.

In f73b3cf, I've taken the first sentence for talking about urgency=2, and used async JavaScript file as an example for urgency=4. 49ea38b adds reference to #merging in respect to urgencies 1 and 5.

draft-kazuho-httpbis-priority.md
kazuho: Thank you for the suggestion.

On one hand I agree, though on the other hand I would be hesitant to use prefetch as an example without talking about reprioritization in general.

Maybe I'd argue that "ordinary web browsing" include all the tasks that are run in the background, while the user uses the web browser for "ordinary web browsing."

Can I confirm you feel strongly about this? Assuming that you do, does changing "ordinary web browsing" to "ordinary web browser use-case" mitigate your concerns?

kazuho:
@rmarx Thank you for the thorough review. I think I've either applied your changes, made adjustments, or have argued against, in response to all your comments.

Would you mind taking a look at them?

draft-kazuho-httpbis-priority.md
rmarx: I think the changes in https://github.com/kazuho/draft-kazuho-httpbis-priority/commit/ebdf06ac255c5cb79b4db6f4fba4da20d2638e77 are (partially) incorrect or can be read that way.

CSS linked in the body will also block rendering, it's not just CSS in the head. The change can be read as though only CSS in the head will block rendering.

I feel for this one we should just stick to your original text, which was correct, just not exhaustive (though, as I've said in other places, I don't think we need to be that exhaustive in the text).

draft-kazuho-httpbis-priority.md
rmarx: Added feedback inline in the first commit, second one looks good!

draft-kazuho-httpbis-priority.md
rmarx: I do indeed feel a bit strongly about this, as I don't think a "software update" makes sense in terms of prioritization on one connection... 

As I see it, software updates would typically happen on a separate connection, not on the same QUIC connection used to load the webpage. As such, you would be talking about cross-connection prioritization/congestion control, which imo is a higher-level issue that doesn't really seem to fit in this text. Unless you envision me downloading an update for Google Chrome while also using google calendar and both endpoints are coalesced onto the same connection or something, but that seems a bit of a stretch and not really the -main- use case here.

Looking at the mail thread where @ianswett and @pmeenan mention the background use cases, it's a bit unclear to me how they see this concretely. They also mention background **uploads**, which also seems a bit weird. Overall, it seems like they envision a browser-wide (or even OS-wide) prioritization setup, managing the full bandwidth of everything combined. In that case, background obviously makes more sense, but again: doesn't fit in this draft? Maybe they can weigh in here? 


On the other hand, I'm struggling to find good uses for "background" that aren't prefetch. Service Worker (cache) updates are one. I was also thinking about "blog post comments" or "social media plugins" but those are probably best at level 4. 

Overall, as I feel we still need a bit of an updated discussion on re-prioritization in the text down the line anyway, we can/should keep the prefetch use case as the example of background.

draft-kazuho-httpbis-priority.md
kazuho: Thank you for your thoughts. Yeah it's hard to strike the best balance between conciseness and clearness.

For this, I think we can simplify. Went with 99eeef4.

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,103 @@ Unknown parameters MUST be ignored.
...
+|               6 | background ({{background}})     |
+{: #urgencies title="Urgencies"}
 
-The default value is `document`.
+The value is encoded as an sh-integer.  The default value is zero.
 
-A server SHOULD transmit HTTP responses in the order of their urgency:
-`blocking` first, followed by `document`, followed by `non-blocking`.
+A server SHOULD transmit HTTP responses in the order of their urgency values.
+The lesser the value is, the higher the precedence is.
```
```suggestion
The lower the value, the higher the precedence.
```

draft-kazuho-httpbis-priority.md
kazuho: I get your point.

While I do not think referring to software updates necessarily lead to the conclusion that we are talking about cross-connection prioritization (they could be coalesced onto another connection that goes to Google), I think you are correct in pointing out that it would be rare to see software update running on a connection that you'd be using for web browsing. Because software update only connects to a specific server.

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,103 @@ Unknown parameters MUST be ignored.
...
-priority = urgency=blocking
+priority = urgency=-1
 ~~~
 
+The definition of the urgencies and their expected use-case are described below.
+Endpoints SHOULD respect the definition of the values when assigning urgencies.
+
+### blocking
+
+The blocking urgency indicates that the response prevents other responses from
```
```suggestion
The blocking urgency (value -1) indicates that the response prevents other responses from
```

As an editorial suggestion I'd move the value to this sentence as shown here , and then lose the following standalone sentence.

kazuho:
@rmarx Thank you once again for your thorough review. The text has become so much better.

To me it seems that the only remaining issue is https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/34#discussion_r302464594. While I agree that it needs to be resolved, I am leaning towards the idea of postponing the discussion on this topic to a separate issue.

The reason is because I think the other parts of the PR is ready, and because this PR is going to block us making progress on the rest of the document, unless we merge this (because this is a big change).

@rmarx @LPardue Thoughts?

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,103 @@ Unknown parameters MUST be ignored.
...
+document, the request for that HTML is given the document urgency.  When that
+HTML document uses a custom font, the request for that custom font SHOULD also
+be given the document urgency.  This is because the availablity of the custom
+font is likely a precondition for the user to use that portion of the HTML
+document, which is to be rendered by that font.
+
+### non-blocking
+
+The non-blocking urgency indicates that the response does not prevent the client
+from using the document even though the response is being incorporated to or
```
```suggestion
from using the document even though the response is being incorporated into or
```

or pick something other than `incorporate to`

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,103 @@ Unknown parameters MUST be ignored.
...
+For example, inline images (i.e., images being fetched and displayed as part of
+the document) are visually important elements of an HTML document.  As such,
+users will typically not be prevented from using the document, at least to some
+degree, before any or all of these images are loaded.  Display of those images
+are thus considered to be an improvement for visual clients rather than a
+prerequisite for all user agents.  Therefore, such images will be given the
+non-blocking urgency.
+
+Values between 1 and 5 are used to represent this urgency, to provide wiggle
+room for endpoints to give some responses more or less precedence than others
```
LPardue: I love the phrase 'wiggle room' but maybe flexibility is a more well-understood term.

draft-kazuho-httpbis-priority.md
```diff
@@ -112,33 +112,103 @@ Unknown parameters MUST be ignored.
...
+
+For example, inline images (i.e., images being fetched and displayed as part of
+the document) are visually important elements of an HTML document.  As such,
+users will typically not be prevented from using the document, at least to some
+degree, before any or all of these images are loaded.  Display of those images
+are thus considered to be an improvement for visual clients rather than a
+prerequisite for all user agents.  Therefore, such images will be given the
+non-blocking urgency.
+
+Values between 1 and 5 are used to represent this urgency, to provide wiggle
```
LPardue: Did I miss where we state that `1 is high` and `5 is low`?

draft-kazuho-httpbis-priority.md
LPardue: I had been meaning to make a similar point to @rmarx here. As a group we have discussed some use cases that are interested but the theme of this proposal is grounded in a web browser activity. Not all network activity from a browser directly related to the navigation or user-facing interaction. While some of these activities might get squeezed into the same connection - coalescing certainly might make that happen, perhaps multipath too? - it seems like somethign on the fringe of this spec's primary purpose.

LPardue (commented):
I've left some editorial feedback as I am a bit limited for a substantial design review.

LPardue:
> To me it seems that the only remaining issue is #34 (comment). While I agree that it needs to be resolved, I am leaning towards the idea of postponing the discussion on this topic to a separate issue.

Agree to kick that can down the road. If we want to change urgency levels, we need to land that ASAP so people can get familiar with it.

draft-kazuho-httpbis-priority.md
kazuho: Makes sense as we have a table that lists the numbers. Applied in 974747c.

draft-kazuho-httpbis-priority.md
kazuho: :+1: Done in 1ec9725.

draft-kazuho-httpbis-priority.md
kazuho: We have the general rule stated; i.e., "The lower the value, the higher the precedence."

That said, I have added 0b516b9 that stresses the meanings of the values.

kazuho:
@LPardue Thank you for your review. With commits up to 0b516b9, I think I have addressed all your concerns (with the description of the "background" urgency being the only exception).

draft-kazuho-httpbis-priority.md
```diff
@@ -257,27 +322,24 @@ end-to-end rather than hop-by-hop.
...
 
 However, that does not mean that the prioritization scheme would forever be
 stuck to the three levels.  The design provides extensibility.  If deemed
-necessary, it would be possible to divide any of the three urgency levels into
-sub-levels by defining a new parameter.  As an example, a server could assign
-an `importance` parameter to the priority of each image that it provides, so
-that an intermediary could prioritize certain images above others.  Or, a
-graphical user-agent could send a `visible` parameter to indicate if the
-resource being requested is within the viewport.
+necessary, it would be possible to subdivide any of the three urgency levels
```
LPardue: there are more than three urgency levels now

kazuho: Nice catch! Fixed in f98aa72.

LPardue (approved):

draft-kazuho-httpbis-priority.md
rmarx: Nice and clear :)

rmarx:
Just created https://github.com/kazuho/draft-kazuho-httpbis-priority/issues/36, so for me it's fine to merge this PR!

kazuho:
@LPardue @rmarx Merged. Thank you!!!

PR #37, opened 2019-07-15
Alternative text for the background use case

### LINKED ISSUE #36: Specify proper use case for background level
As discussed in this small thread https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/34#discussion_r302464594, the background example use case of "downloading an update" doesn't really seem to fit the typical web browsing use case.

Other options: Prefetch (but that requires discussing re-prioritization), Service Worker-based updates, things like blog comments and social media plugins. 

Welcoming input from @ianswett and @pmeenan on this one as well.

rmarx:
Fixes #36. 
Not entirely sure of the wording and how detailed the description needs to be.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -203,7 +203,7 @@ be postponed without having an impact on using other responses.
 
 As an example, the download of a large file in a web browser would be assigned
 the background urgency so it would not impact further page loads on the same
-domain.
+connection.
 
 ## progressive
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -201,8 +201,9 @@ used.
 The background urgency (value 6) is used for responses of which the delivery can
 be postponed without having an impact on using other responses.
 
-As an example, a software update running in the background would be assigned the
-background urgency.
+As an example, the download of a large file in a web browser would be assigned
+the background urgency so it would not impact further page loads on the same
+domain.
 
 ## progressive
 


draft-kazuho-httpbis-priority.md
```diff
@@ -201,8 +201,9 @@ used.
 The background urgency (value 6) is used for responses of which the delivery can
 be postponed without having an impact on using other responses.
 
-As an example, a software update running in the background would be assigned the
-background urgency.
+As an example, the download of a large file in a web browser would be assigned
+the background urgency so it would not impact further page loads on the same
+domain.
```
LPardue: Maybe there is a more correct phrase than "same domain"?

rmarx: I agree. Options:
- same connection
- same host
- to the same origin

kazuho: I think "connection" would be fine.

The concern here might be that "connection" implies hop-by-hop rather than end-to-end, but I would not be too worried, considering that this is an example. If we want to describe how the priority header field can be used in hops other than the one that originates from the end client, I think there'd be a place better than here.

kazuho (commented):
Thanks @rmarx. Looks much better. I think the PR is ready for merge modulo the point raised by @LPardue.

draft-kazuho-httpbis-priority.md
LPardue: same connection is most appropriate I think.

rmarx:
Resolved outstanding point, should be ready for merge.

LPardue (approved):

kazuho (approved):

kazuho:
Merged. Thank you!!

PR #38, opened 2019-07-15
use less browser-specific terms

kazuho:
This PR attempts to describe the urgency levels using terms that are less specific to web browsers or how the web browsers process the responses.

* I think "default" is the best term to be used in place of "document".
* I also like "supplementary", because I think it implies the difference between custom fonts (which is given "default" urgency) and images (which are "supplementary").
* I am not sure if "prerequisite" sounds good, though I think it's logically correct.

@rmarx @pmeenan @LPardue Thoughts?

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -114,12 +114,12 @@ Unknown parameters MUST be ignored.
 
 The `urgency` parameter takes an integer between -1 and 6 as shown below:
 
-| Urgency         | Definition                      |
-|----------------:|:--------------------------------|
-|              -1 | blocking ({{blocking}})         |
-|               0 | document ({{document}})         |
-| between 1 and 5 | non-blocking ({{non-blocking}}) |
-|               6 | background ({{background}})     |
+| Urgency         | Definition                        |
+|----------------:|:----------------------------------|
+|              -1 | prerequisite ({{prerequisite}})   |
+|               0 | default ({{default}})             |
+| between 1 and 5 | supplementary ({{supplementary}}) |
+|               6 | background ({{background}})       |
 {: #urgencies title="Urgencies"}
 
 The value is encoded as an sh-integer.  The default value is zero.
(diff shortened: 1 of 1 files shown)


rmarx:
I'm fine with default and supplementary, but am indeed not a fan of "prerequisite".

Using a thesaurus, some options that I would find preferable:
"essential", "core", "cardinal", "critical" (though that might conflate with the meaning of the "critical path"), or, nicely symmetric with supplementary: "necessary".

I wouldn't bikeshed this too much though. I'm fine with whatever as long as it doesn't change existing semantics, so "prerequisite" is ok as well.

draft-kazuho-httpbis-priority.md
```diff
@@ -141,51 +141,52 @@ priority = urgency=-1
...
+to be rendered by that font.
 
-### non-blocking
+### supplementary
 
-The non-blocking urgency indicates that the response does not prevent the client
-from using the document even though the response is being incorporated into or
-referred to by the document.
+The supplementary urgency indicates a response that is helpful to the client
+using a composition of HTTP responses, even though the response itself is not
```
LPardue: nit, you previously made an editorial pass to remove `HTTP`
```suggestion
using a composition of responses, even though the response itself is not
```

LPardue (approved):
found a nit but otherwise LGTM.

I think the prerequisite term has nice interplay with guidance for a server to use request generation order to determine how to prioritise responses.

LPardue:
Although perhaps technically you learn the dependency during the document load and parse (not before or after it) so maybe we use the prefix "peri" i.e. perirequisite. :laughing:

draft-kazuho-httpbis-priority.md
kazuho: Thanks! Applied.

kazuho:
> Although perhaps technically you learn the dependency during the document load and parse (not before or after it) so maybe we use the prefix "peri" i.e. perirequisite. 😆

One of the things I really enjoy about coauthoring with @LPardue is that I can learn a lot about English and English culture (maybe British to some extent, but I do not know). It's a nice spice while working on a dry technical writing.

That said, I think we are fine with "prerequisite" here, because we are talking about "using" the document rather than "processing" it. One of the reasons I prefer making this change is that the new terms more closely aligns to "using" compared to the current terms.

LPardue:
Yeah it works fine for me. I just cant help myself making silly jokes. Let's merge

pmeenan:
LGTM.

kazuho:
@rmarx @LPardue @pmeenan Thank you for your comments. Merged.

Issue #33, opened 2019-07-10
Request/response parameter omission asymmetry


LPardue:
> In my mind, the rules for priority header were quite symmetric but I think this highlights an asymmetry between how they are applied across requests and responses. I think we might want to spin that out into a separate issue, WDYT?

_Originally posted by @LPardue in https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/30_

Expansion of server override behaviour text seemed to highlight something I had not realised. We might want to try and explain this better in the parameter definition area.

kazuho:
FWIW, current text has the following sentence, but I agree that there is no text that specifies what happens when a parameter is missing.
> When used, the parameters found in the response header field overrides those specified by the client.

Let me elaborate why I think the values of the unused parameters should be the ones that were specified by the client.

Consider the case where a server is serving a baseline JPEG (i.e. non-progressive JPEG file). It might want to send `progressive=?0`, because it thinks that there's no benefit in sending the image in small chunks. This HTTP response can be cached.

Requests that fetch this JPEG image might contain different urgency levels. When it is used as part of a HTML document, it would likely be `urgency=non-blocking`. But if the user is saving the image to disk as a background operation, it might have `urgency=background` (assuming that we define "background").

Or else, we might define an extension called "visibility." I'd assume that we'd want the H2/H3 terminator to respect the value of the client, unless the server explicitly specifies a visibility value.

To generalize, it is my view that (if not all, at least some) parameters are orthogonal to each other. A server specifying one parameter does not mean that the server is specifying others. Hence the parameters unspecified by the server should default to the client-provided value.

LPardue:
Ok so I dread to suggest it but I think what might help is a section on Priority information lifetime. I.e. before a request is generated a resource has  default priority. This can be initialised to non-defauls (using an information tuple) and then later updated (but not requiring a tuple, aka piecemeal). Let me sleep on that idea for a bit.

rmarx:
I agree with Kazuho that the client-provided parameters should be kept if the server overrides just some of them. 

I am also in favor of more clarifications in general, so the Priority Information lifetime gets a thumbs-up from me!

LPardue:
Ok. I'll wait a little and let some of the other discussion items settle down a bit and then make a start on this.

LPardue:
I think this has been overtaken, in whole or part, by other events. So I'm closing and will reopen if needs be later.

PR #40, opened 2019-07-17
Expand collaboration and contrast to H2

LPardue:
This goes someway to addressing #33.

I deviated from my original plan and focused more on what H2 refers to as a prioritization process. I highlight that H2 only gives clients explicit input to the process, and attempt to describe that a more collaborative process could improve the overall usability for the client. (_aside: collaboration replaces the use of override_).

I also try to explain and delineate between the initial priority that is possible with the Priority header, and subsequent reprioritization that might require a frame or whatever.

Finally, we go back to the merging example but call it collaboration.

Its not perfect, we might want to chop it up and only pull out the useful bits. Elsewhere some opinions have been shared on breaking up into separate documents so we might want to put some of this text elsewhere.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -236,10 +236,10 @@ priority = urgency=3, progressive=?1
 # Prioritization Process and Lifecycle
 
 HTTP/2 alludes to a prioritization process and does not specify many details
-about it. This might go some way to explain the different perspectives about
-prioritization that arise from different operators and use cases. The design
-also mixes concerns about prioritization between requests, streams and
-connections.
+about it. The prioritization scheme mixes concerns about requests, streams and
+connections, which complicates understanding. These factors might go some way to
+explain the different perspectives about prioritization that arise from
+different operators and use cases.
 
 HTTP/2's frame-based prioritization places an emphasis on the client's role in
 determining the priority of a response. {{?RFC7540}} Section 5.3 describes
@@ -265,7 +265,7 @@ generation, the server can possibly determine that the client's predicted
 prioritization was incorrect (perhaps sub-optimal) for how the selected
 representation will be used. An HTTP/2 server might use this information,
 together with other inputs, to send the response at a different effective
-priority than the client suggested. However, it has no explicit means on which
+priority than the client suggested. However, it has no explicit means by which
 to present such a finding to the client.
 
 The frame-based prioritization process weakens further when intermediaries are
@@ -277,12 +277,12 @@ A collaborative model is presented that attempts to abstract priority and
 collaboration from concerns about HTTP version and active connections.
 
 Resources have an initial priority that begins with a neutral value. Requests
-and responses permit collaborative changes to the resources initial priority
+and responses permit collaborative changes to the resource's initial priority
 information. A client attaches suggested priority information to the request,
 omission activates a default priority. A server can (if supported) attach
 priority information to the response, omission might be an acceptance of the
-client offer or or implicit ignore. Once a request or response is in flight, an
-endpoint cannot modify the initial priority, further changes are classed as
+client offer or an implicit ignore. Once a request or response is in flight, an
+endpoint cannot modify the initial priority, any further change is classed as a
 reprioritization that requires a non-request or non-response carriage mechanism
 (e.g. a frame).
 
@@ -291,13 +291,13 @@ prioritization signal for servers.
 
 ## Explicit Client and Server Collaboration {#collaboration}
 
-The "Priority" header field can be attached to responses in order to provide an
-explicit signal that allows servers collaborate to collaborate on the
-prioritization process. A server sends new priority parameter values or confirms
-the client suggestion. This process works well in the case of intermediaries
-because the client suggested priorities are passed to the origin, giving it the
-opportunity to improve the priority information before the intermediary
-allocates resources to the response.
+The "Priority" header field can be attached to responses, it is an explicit
+signal that allows a server to collaborate in the prioritization process by
+sending new priority parameter values or confirming the client suggestion. When
+an intermediary is present, the client-suggested priority is passed to the
+origin server, which gives it the opportunity to improve the priority
+information. The intermediary can use the origin server response to decide how
+to best prioritize serving the response.
 
 For example, a client using an HTML document discovers several inline images and
 requests them with a low urgency via an intermediary. The origin server has
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -233,71 +233,17 @@ set to `3` and the progressive parameter set to `1`.
 priority = urgency=3, progressive=?1
 ~~~
 
-# Prioritization Process and Lifecycle
+# Merging Client- and Server-Driven Parameters {#merging}
 
-HTTP/2 alludes to a prioritization process and does not specify many details
-about it. The prioritization scheme mixes concerns about requests, streams and
-connections, which complicates understanding. These factors might go some way to
-explain the different perspectives about prioritization that arise from
-different operators and use cases.
+The "Priority" header field is an end-to-end signal ({{end-to-end}}) that allows
+endpoints to collaborate on priorities ({{collaboration}}).
 
-HTTP/2's frame-based prioritization places an emphasis on the client's role in
-determining the priority of a response. {{?RFC7540}} Section 5.3 describes
-one key characteristic:
-
-> Explicitly setting the priority for a stream is input to a
-  prioritization process.  It does not guarantee any particular
-  processing or transmission order for the stream relative to any other
-  stream.  An endpoint cannot force a peer to process concurrent
-  streams in a particular order using priority.  Expressing priority is
-  therefore only a suggestion.
-
-In practice, this approach is quite one-sided and puts a burden on the endpoint
-with the most limited amount of information. A client processing HTML and
-generating requests generally has only a URL and a type (determined by an HTML
-element or other mechanism) on which to make a prioritization prediction.
-
-A server typically has more information about the resource than the client,
-including the context around it (such as linked representations). It uses local
-knowledge, with input from the client, to generate a single selected
-representation ({{?RFC7231}}, Section 3) and respond. During response
-generation, the server can possibly determine that the client's predicted
-prioritization was incorrect (perhaps sub-optimal) for how the selected
-representation will be used. An HTTP/2 server might use this information,
-together with other inputs, to send the response at a different effective
-priority than the client suggested. However, it has no explicit means by which
-to present such a finding to the client.
-
-The frame-based prioritization process weakens further when intermediaries are
-considered because only the client is permitted to provide an explicit signal.
-
-## A Collaborative Model
-
-A collaborative model is presented that attempts to abstract priority and
-collaboration from concerns about HTTP version and active connections.
-
-Resources have an initial priority that begins with a neutral value. Requests
-and responses permit collaborative changes to the resource's initial priority
-information. A client attaches suggested priority information to the request,
-omission activates a default priority. A server can (if supported) attach
-priority information to the response, omission might be an acceptance of the
-client offer or an implicit ignore. Once a request or response is in flight, an
-endpoint cannot modify the initial priority, any further change is classed as a
-reprioritization that requires a non-request or non-response carriage mechanism
-(e.g. a frame).
-
-As explained above, in this model HTTP/2 does not provide an explicit
-prioritization signal for servers.
-
-## Explicit Client and Server Collaboration {#collaboration}
-
-The "Priority" header field can be attached to responses, it is an explicit
-signal that allows a server to collaborate in the prioritization process by
-sending new priority parameter values or confirming the client suggestion. When
-an intermediary is present, the client-suggested priority is passed to the
-origin server, which gives it the opportunity to improve the priority
-information. The intermediary can then use the origin server response to decide how
-to best prioritize serving the response.
+Client-driven parameters are sent in requests and server-driven parameters are
+sent in responses. When an intermediary is present, the client's suggested
+priority is passed to the origin server, which gives it the opportunity to
+improve the priority information. The intermediary can then use the origin
+server response in order to decide how to best prioritize the response that it
+sends.
 
 For example, a client using an HTML document discovers several inline images and
 requests them with a low urgency via an intermediary. The origin server has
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+  streams in a particular order using priority.  Expressing priority is
+  therefore only a suggestion.
+
+In practice, this approach is quite one-sided and puts a burden on the endpoint
+with the most limited amount of information. A client processing HTML and
+generating requests generally has only a URL and a type (determined by an HTML
+element or other mechanism) on which to make a prioritization prediction.
+
+A server typically has more information about the resource than the client,
+including the context around it (such as linked representations). It uses local
```
kazuho: To me it seems that the comparison regarding the information that the endpoints have is a bit misleading.

Type (how a resource is used) is something that the client would always have, but something that a server "might" have. OTOH, it is true that the representation is only known to the server.

Maybe it's better to clarify that letting the server decide how to prioritize leads to better results, because the information that the both endpoints have can be "combined".

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+representation ({{?RFC7231}}, Section 3) and respond. During response
+generation, the server can possibly determine that the client's predicted
+prioritization was incorrect (perhaps sub-optimal) for how the selected
+representation will be used. An HTTP/2 server might use this information,
+together with other inputs, to send the response at a different effective
+priority than the client suggested. However, it has no explicit means on which
+to present such a finding to the client.
+
+The frame-based prioritization process weakens further when intermediaries are
+considered because only the client is permitted to provide an explicit signal.
```
kazuho: The text up to here seems to be more focused about explaining the problem of H2 rather than describing the design of the Priority header. Therefore, I might suggest making it a subsection of "Introduction" (or "Considerations").

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+
+## A Collaborative Model
+
+A collaborative model is presented that attempts to abstract priority and
+collaboration from concerns about HTTP version and active connections.
+
+Resources have an initial priority that begins with a neutral value. Requests
```
kazuho: To to it seems that it is from this line that the text actually talks about "Lifecycle."

I like the approach of talking about the lifecycle and how the request and response header field are being merged. To rephrase, I'm in favor of keeping the lines below this paragraph at the current position.

kazuho (commented):
This is a great write up, as clarifies the rationale about the design.

But I might suggest restructuring, as it is more about the background rather than the proposed design itself.

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
-
-By using the "Priority" response header, a server can override the
-prioritization hints provided by the client.  When used, the parameters found
-in the response header field overrides those specified by the client.
-
-For example, when the client sends an HTTP request with
+# Prioritization Process and Lifecycle
+
+HTTP/2 alludes to a prioritization process and does not specify many details
+about it. This might go some way to explain the different perspectives about
```
rmarx: what is the "This" in this sentence referring to?

draft-kazuho-httpbis-priority.md
rmarx: Yes, I also don't really agree with how you're explaining it here.

While it's correct in principle, in practice the server typically doesn't really have that extended context available in an easy to use way or at least doesn't make appropriate use of it in most current deployments. Next to that, clients have more information than just URL and type, e.g., where the resource is in the HTML plays a large role, as well as how the internal implementation works.

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+
+A server typically has more information about the resource than the client,
+including the context around it (such as linked representations). It uses local
+knowledge, with input from the client, to generate a single selected
+representation ({{?RFC7231}}, Section 3) and respond. During response
+generation, the server can possibly determine that the client's predicted
+prioritization was incorrect (perhaps sub-optimal) for how the selected
+representation will be used. An HTTP/2 server might use this information,
+together with other inputs, to send the response at a different effective
+priority than the client suggested. However, it has no explicit means on which
```
rmarx: > means on which

means *by* which?

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+element or other mechanism) on which to make a prioritization prediction.
+
+A server typically has more information about the resource than the client,
+including the context around it (such as linked representations). It uses local
+knowledge, with input from the client, to generate a single selected
+representation ({{?RFC7231}}, Section 3) and respond. During response
+generation, the server can possibly determine that the client's predicted
+prioritization was incorrect (perhaps sub-optimal) for how the selected
+representation will be used. An HTTP/2 server might use this information,
+together with other inputs, to send the response at a different effective
```
rmarx: maybe give an example of "other inputs" (e.g., manual annotations by the developers)

draft-kazuho-httpbis-priority.md
rmarx: I'm not sure I follow you on this. I've seen it said in other places that priorities are inherently hop-by-hop (I think Tom Bergan on the mailing list said this?). That seems counter to what you're claiming here.

rmarx: I agree. If you're going with a "this is how it's done in H2 and this is what's wrong with it and this is how this proposal fixes it"-setup, I would make this more explicit (maybe including an explicit list of all pain points and coming back to that in the "A collaborative model" section?)

For example, the part:
> As explained above, in this model HTTP/2 does not provide an explicit
prioritization signal for servers.

Could be moved up to this section and then referenced to in the later section.

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+The frame-based prioritization process weakens further when intermediaries are
+considered because only the client is permitted to provide an explicit signal.
+
+## A Collaborative Model
+
+A collaborative model is presented that attempts to abstract priority and
+collaboration from concerns about HTTP version and active connections.
+
+Resources have an initial priority that begins with a neutral value. Requests
+and responses permit collaborative changes to the resources initial priority
```
rmarx: > resources

resource's or resources'

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+information. A client attaches suggested priority information to the request,
+omission activates a default priority. A server can (if supported) attach
+priority information to the response, omission might be an acceptance of the
+client offer or or implicit ignore. Once a request or response is in flight, an
```
rmarx: > or or 

or an

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+endpoint cannot modify the initial priority, further changes are classed as
+reprioritization that requires a non-request or non-response carriage mechanism
```
rmarx: > reprioritization that requires

reprioritizations that require

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+(e.g. a frame).
+
+As explained above, in this model HTTP/2 does not provide an explicit
+prioritization signal for servers.
+
+## Explicit Client and Server Collaboration {#collaboration}
+
+The "Priority" header field can be attached to responses in order to provide an
+explicit signal that allows servers collaborate to collaborate on the
```
rmarx: > allows servers collaborate to collaborate

allows servers to collaborate

Also, this sentence doesn't seem to make much sense with what comes later. You talk about servers + clients and origins + intermediates. As such, maybe better to say here "allows endpoints to collaborate"? (although endpoint is not a perfect term for an intermediary...)

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+prioritization process. A server sends new priority parameter values or confirms
+the client suggestion. This process works well in the case of intermediaries
```
rmarx: > the client suggestion

the client's suggestion

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+because the client suggested priorities are passed to the origin, giving it the
```
rmarx: > to the origin

to the origin server

draft-kazuho-httpbis-priority.md
```diff
@@ -233,18 +233,78 @@ set to `3` and the progressive parameter set to `1`.
...
+opportunity to improve the priority information before the intermediary
+allocates resources to the response.
+
+For example, a client using an HTML document discovers several inline images and
+requests them with a low urgency via an intermediary. The origin server has
+knowledge that one of the images heavily affects the usage of the HTML document
```
rmarx, line 250: > heavily affects the usage of the HTML document

heavily impacts the end user's experience

LPardue:
Thank you @kazuho and @rmarx for the feedback. I'll try to fix the nits first and then restructure things. But also feel free to suggest text to be cut. I'm happy to land changes that improve upon the current document and leave text that is too wooly (or controversial) on the side for a rainy day.

draft-kazuho-httpbis-priority.md
LPardue: I stuck with use of HTML in order to align with @kazuho terminology.

I think we have an overarching appreciation of user experience but focus on individual resource usage in specific examples.

LPardue (commented):
@rmarx I tried to address your comment in d7cdd2de7c493d4c5d55085c8017cf5045a5f8fa.

A follow up PR will shuffle text around a bit more.

draft-kazuho-httpbis-priority.md
```diff
@@ -291,13 +291,13 @@ prioritization signal for servers.
...
-the client suggestion. This process works well in the case of intermediaries
-because the client suggested priorities are passed to the origin, giving it the
-opportunity to improve the priority information before the intermediary
-allocates resources to the response.
+The "Priority" header field can be attached to responses, it is an explicit
+signal that allows a server to collaborate in the prioritization process by
+sending new priority parameter values or confirming the client suggestion. When
+an intermediary is present, the client-suggested priority is passed to the
+origin server, which gives it the opportunity to improve the priority
+information. The intermediary can use the origin server response to decide how
```
rmarx:
```suggestion
information. The intermediary can then use the origin server response to decide how
```

rmarx (changes requested):

draft-kazuho-httpbis-priority.md
rmarx: While I am all in favor of consistency, I don't think the sentence as it stands makes much sense semantically... Maybe changing to "the way the HTML document is perceived by the end user" could help? or "the practical usage of the HTML document"?

draft-kazuho-httpbis-priority.md
LPardue: fixed in a different way, maintaining non-plural term.

LPardue:
With the latest commit, I restored the Merging section so it looks very similar to the version on master. I moved the prose bits to considerations and attempted to address your comments about who has more information at the same time.

Apologies if this makes re-review difficult but I think it is a step in correct direction that was needed.

draft-kazuho-httpbis-priority.md
```diff
@@ -99,7 +99,7 @@ The Priority HTTP header field can appear in requests and responses. A client
 uses it to specify the priority of the response. A server uses it to inform
 the client that the priority was overwritten. An intermediary can use the
 Priority information from client requests and server responses to correct or
-amend the precedence to suit it (see {{merging}}).
+amend the precedence to suit it (see {{collaboration}}).
```
kazuho: I think that the reference should point to "merging", as htat that is the section that defines the procedure of "amending" the precedence.

draft-kazuho-httpbis-priority.md
```diff
@@ -235,16 +235,22 @@ priority = urgency=3, progressive=?1
...
 
-By using the "Priority" response header, a server can override the
-prioritization hints provided by the client.  When used, the parameters found
-in the response header field overrides those specified by the client.
+Client-driven parameters are sent in requests and server-driven parameters are
+sent in responses. When an intermediary is present, the client's suggested
+priority is passed to the origin server, which gives it the opportunity to
+improve the priority information. The intermediary can then use the origin
+server response in order to decide how to best prioritize the response that it
+sends.
```
kazuho, line 246: I think you've removed the sentence that specifies how the client- and server-provided hints are merged; i.e. "When used, the parameters found in the response header field overrides those specified by the client."

Also, do we need to talk about a caching intermediary? The proposed text suggests that the request traverses to the origin server, but does not clarify what happens when the intermediary has the response cached.

draft-kazuho-httpbis-priority.md
```diff
@@ -235,16 +235,22 @@ priority = urgency=3, progressive=?1
...
-dependencies is typically best known to the server.
+The "Priority" header field is an end-to-end signal ({{end-to-end}}) that allows
+endpoints to collaborate on priorities ({{collaboration}}).
 
-By using the "Priority" response header, a server can override the
-prioritization hints provided by the client.  When used, the parameters found
-in the response header field overrides those specified by the client.
+Client-driven parameters are sent in requests and server-driven parameters are
+sent in responses. When an intermediary is present, the client's suggested
+priority is passed to the origin server, which gives it the opportunity to
```
kazuho, line 243: > When an intermediary is present, the client's suggested priority is passed to the origin server

I am not sure if it this is correct, considering the fact that a cached intermediary might respond to the client without contacting the origin. If we are to suggest something like this, I think that we should also talk about what happens when a cached response is returned.

draft-kazuho-httpbis-priority.md
```diff
@@ -302,7 +307,71 @@ extension element (see {{!RFC7540}}, Section 5.5).
...
+
+> Explicitly setting the priority for a stream is input to a
+  prioritization process.  It does not guarantee any particular
+  processing or transmission order for the stream relative to any other
+  stream.  An endpoint cannot force a peer to process concurrent
+  streams in a particular order using priority.  Expressing priority is
+  therefore only a suggestion.
+
+The client is not alone in understanding the best way to prioritize a request. A
+server's prioritization decision is driven by a number of factors such as the
```
kazuho: Maybe s/is/can be/?

I suggest this because we are talking about possibilities.

draft-kazuho-httpbis-priority.md
```diff
@@ -302,7 +307,71 @@ extension element (see {{!RFC7540}}, Section 5.5).
...
+selected representation ({{?RFC7231}}, Section 3), data locality, resource
+contention, load balancing across multiple clients, application-specific
+annotations etc. When it generates responses it might determine that the
+client's predicted prioritization was incorrect for how the selected
+representation will be used. Rigidly following the client's suggestion could
```
kazuho, line 337: I am not sure if we should talk about a server detecting the client's prediction being incorrect.

The reason is because that would not be the case, especially when an intermediary uses a response it has cached, that has the priority response header. Such an intermediary just obeys to the origin's instruction, rather than determining something as wrong.

To put differently, the priority request and response headers are unilateral declarations.

draft-kazuho-httpbis-priority.md
```diff
@@ -302,7 +307,71 @@ extension element (see {{!RFC7540}}, Section 5.5).
...
+lead to a sub-optimal user experience, and so to avoid this, the server can send
+the response at a different effective priority than the client suggestion.
+
+The prioritization process is a collaborative activity but there are no explicit
+means by which a server can present its information. Attempting to infer reasons
+about prioritization from implicit signals is impractical and error prone due to
+the number of factors involved.
```
kazuho, line 344: Re "attempting to infer reasons about prioritization", I think we need to be clear that the subject is an intermediary. Because sending the priority response header to the client does not affect how the responses are prioritized.

draft-kazuho-httpbis-priority.md
```diff
@@ -302,7 +307,71 @@ extension element (see {{!RFC7540}}, Section 5.5).
...
+omission activates a default priority. A server can (if supported) attach
+priority information to the response, omission might be an acceptance of the
+client offer or an implicit ignore. Once a request or response is in flight, an
+endpoint cannot modify the initial priority, any further change is classed as a
+reprioritization that requires a non-request or non-response carriage mechanism
+(e.g. a frame).
+
+Today's HTTP/2 priority information is exchanged as frames and restricted to
+clients. In order to support the new process, changes to frames and their
+handling are required. Extending behavior at that this level has proven hard
```
kazuho: proven hard "to"?

kazuho (commented):
Thank you for the changes.

I like the new structure, though I think we need to discuss before merging the PR. I've left my tentative comments. PTAL.

draft-kazuho-httpbis-priority.md
LPardue: Good spot. I'll take a stab at this.

draft-kazuho-httpbis-priority.md
LPardue: Braking this into two separate comments:

1) In the re-re-shuffle the description of merging is inexact, I will fix this

2) I think the point about caching intermediaries relates to your previous comment. Although caching intermediaries are discussed in considerations, there are no examples of them. Which makes it easy to miss this capability. I'll have a go, which I'll probably get wrong. Feel free to suggest one too.

draft-kazuho-httpbis-priority.md
LPardue: I appreciate the point you are making in this and the next comment. I think the general problem with the text as it stands is that it tries to describe how HTTP/2 works today, *and* make the case for an explicit server-driven priority signal.

To the first point I was attempting to describe a case we have today in HTTP/2 where a normal server gets a a request that is obviously a dumb prioritisation e.g. let's pick on Spartan-Edge. By ignoring the suggestion, we can improve page load time.

However, for the above case, no explicit signal is required and I think therefore that the argument is not strong (as you point out "sending the priority response header to the client does not affect how the responses are prioritized."). 

> Such an intermediary just obeys to the origin's instruction, rather than determining something as wrong.

I think this is the root to one of the problems with this PR. Is it your opinion that collaboration of initial priority involves only  the client and the origin? And that the intermediary cannot override parameters, all it can do is use the priority information in combination with other factors in order to manage the resource prioritization on the HTTP/2 hop to the client.

LPardue: I think these were points that @rmarx was trying to make in a previous review too.

LPardue:
This has rotten, both because the discussion has moved on and also due to document changes. For the purposes of clarity, I'm going to close the PR and issue, and reopen if later assesment determines there is still a problem/benefit.

PR #41, opened 2019-07-18
explain that prioritized queuing of requests is difficult with the H2 scheme.

### LINKED ISSUE #39: Talk about using urgency for limiting backend concurrency
An intermediary might have a queue that enqueues the requests received from the client, so that the number of requests that it issues to the backend can be capped.

One example is the [http2-max-concurrent-requests-per-connection](https://h2o.examp1e.net/configure/http2_directives.html#http2-max-concurrent-requests-per-connection) configuration directive of H2O. It limits the number of requests that H2O issues to the handlers, rather than the number of requests that the client can issue.

When picking a pending request from such a queue, it would be a good idea to consult the urgency value of those requests.

kazuho:
@LPardue @rmarx WDYT?

Closes #39.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -71,7 +71,7 @@ bandwidth among the HTTP responses.  However, the design has shortcomings:
   coordinate client-driven and server-driven priorities.
 * It is hard for a server to determine the processing order of queued requests
   based on their priorities, when the server's internal request concurrency is
-  limited differently than the concurrency of HTTP/2 connections.
+  limited differently than the stream-level concurrency of HTTP/2 connections.
 * The design cannot be ported cleanly to HTTP/3 ({{?I-D.ietf-quic-http}}).  One
   of the primary goals of HTTP/3 is to minimize head-of-line blocking.
   Transmitting the evolving representation of a "prioritization tree" from the
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -69,6 +69,9 @@ bandwidth among the HTTP responses.  However, the design has shortcomings:
 * It does not define a method that can be used by a server to express the
   priority of a response.  Without such a method, intermediaries cannot
   coordinate client-driven and server-driven priorities.
+* It is hard for a server to determine the processing order of queued requests
+  based on their priorities, when the server's internal request concurrency is
+  limited differently than the concurrency of HTTP/2 connections.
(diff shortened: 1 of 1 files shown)


rmarx:
I understand what you're saying because of the conversation in the other issue, but I doubt most other people would. The text seems overly obtuse, but I can't really think of a terse way to state things more clearly atm.

LPardue:
Just so I build my own understanding, are you trying to say something like.

" Internal server concurrency limits can be different from HTTP/2 connection limits. Since the prioritization tree is built in respect of the connection, it is difficult for the server to directly map the tree to internal processes."

kazuho:
> Just so I build my own understanding, are you trying to say something like.
> 
> " Internal server concurrency limits can be different from HTTP/2 connection limits.

Exactly.

> Since the prioritization tree is built in respect of the connection, it is difficult for the server to directly map the tree to internal processes."

There are two ways to cap the concurrency. One is to cap the total number of requests that the server processes concurrently. The other is to cap the number of requests being concurrently processed per each H2 connection.

That latter makes sense - the intent of running as much as 100 concurrent requests per connection is to fulfill the bandwidth. If prioritization is unnecessary, the server can limit the number of requests it processes per each connection to just one (assuming that it can buffer other requests), assuming that the response becomes available immediately once the request is processed.

Or, ignoring the "progressiveness" parameter for a moment, minimum-maximum of the "request *processing* concurrency" required by a H2 terminator that supports the priority header is 8, assuming that the response becomes immediately available once the request is processed. This is because having one pending "response" for each urgency level is sufficient to get the most out of the prioritization scheme (disregarding the progressiveness).

The question is how you could determine the such minimum-maximum in the H2 scheme, and how a server would pick up the request to be processed.

LPardue:
Thank you for the explanation. While valid this seems something hard to articulate concisely and I agree with Robin that even if all the words make sense, their true meaning doesn't shine.

Perhaps you want to make this a sub-bullet of the first one that talks about complexity affecting support levels. I susggest this because it seems like one particular manifestation of the overall problem with trying to apply the flexible nature of the H2 scheme to a server's more rigid resource allocation and other needs.

LPardue:
@kazuho do you want to pursue this change? The conflict comes from the changes I made to add a new section for the reasons replace H2 priorities. That section borrowed a bit of prose from various sources but I don't know if it obviates the change you had in mind here and/or that how you want it to look based on the feedback.

kazuho:
Let's close. We can revisit if necessary.

PR #51, opened 2019-10-18
expand reprioritization section and define frame for H2 and H3

LPardue:
This expands the reprioritization section so that it defines the frames rather than just suggest they are needed. I moved some text around and made the example less specific to a version. However, I refrained from moving the Reprioritization section out of considerations - for now.

The HTTP/3 version of this frame needs the ability to reprioritize a normal response or a pushed one. I chose to add a type bit flag so that the frame can carry a push ID and thus allow the client to reprioritize a push before it starts. The other option is to remove the bit field and only carry a stream ID and restrict the client to only reprioritize the push after it begins.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -316,12 +316,19 @@ field.  This is because an HTTP header field can only be sent as part of an HTTP
 message. Therefore, to support reprioritization, it is necessary to define a
 HTTP-version-dependent mechanism for transmitting the priority parameters.
 
-This document specifies a new frame type for HTTP/2 ({{!RFC7540}}) and HTTP/3
-({{!I-D.ietf-quic-http}}) that is specialized for reprioritization. It
-carries updated priority parameters and is sent on the version-specific control
-stream. The frame references the response to reprioritize based on a
-version-specific identifier; in HTTP/2 this is the Stream ID, in HTTP/3 this is
-either the Stream ID or Push ID.
+This document specifies a new PRIORITY_UPDATE frame type for HTTP/2
+({{!RFC7540}}) and HTTP/3 ({{!I-D.ietf-quic-http}}) that is specialized for
+reprioritization. It carries updated priority parameters and references the
+target of the reprioritization based on a version-specific identifier; in
+HTTP/2 this is the Stream ID, in HTTP/3 this is either the Stream ID or Push ID.
+
+In HTTP/2 and HTTP/3 a request message sent on a stream transitions it into a
+state that prevents the client from sending additional frames on the stream.
+Modifying this behavior requires a semantic change to the protocol, this is
+avoided by restricting the stream on which a PRIORITY_UPDATE frame can be sent.
+In HTTP/2 the frame is on stream zero and in HTTP/3 it is sent on the control
+stream ({{!I-D.ietf-quic-http}}, Section 6.2.1).
+
 
 ## HTTP/2 PRIORITY_UPDATE Frame
 
@@ -329,7 +336,8 @@ The HTTP/2 PRIORITY_UPDATE frame (type=0xF) carries the stream ID of the
 response that is being reprioritized, and the updated priority in ASCII text,
 using the same representation as that of the Priority header field value.
 
-The stream identifier for a PRIORITY_UPDATE frame MUST be zero (0x0).
+The Stream Identifier field ({{!RFC7540}}, Section 4.1) in the PRIORITY_UPDATE
+frame header MUST be zero (0x0).
 
 ~~~ drawing
   0                   1                   2                   3
@@ -351,7 +359,7 @@ The HTTP/3 PRIORITY_UPDATE frame (type=0xF) carries the identifer of the element
 that is being reprioritized, and the updated priority in ASCII text, using the
 same representation as that of the Priority header field value.
 
-The PRIORITY_UPDATE frame is sent on the control stream
+The PRIORITY_UPDATE frame MUST be sent on the control stream
 ({{!I-D.ietf-quic-http}}, Section 6.2.1).
 
 ~~~ drawing
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -302,6 +302,84 @@ receipt, a client that supports header-based prioritization MUST close the
 connection with a protocol error. Non-supporting clients will ignore this
 extension element (see {{!RFC7540}}, Section 5.5).
 
+# Reprioritization
+
+Once a client sends a request with a priority, circumstances might change and
+mean that it is beneficial to change the priority of the response. As an
+example, a web browser might issue a prefetch request for an HTML document with
+the urgency parameter of the Priority request header field set to `background`.
+Then, when the user navigates to the HTML while prefetch is in action, it would
+send a reprioritization frame with the priority field value set to `urgency=0`.
+
+However, a client cannot reprioritize a response by using the Priority header
+field.  This is because an HTTP header field can only be sent as part of an HTTP
+message. Therefore, to support reprioritization, it is necessary to define a
+HTTP-version-dependent mechanism for transmitting the priority parameters.
+
+This document specifies a new frame type for HTTP/2 ({{?RFC7540}}) and HTTP/3
+({{?I-D.draft-ietf-quic-http-23}}) that is specialized for reprioritization. It
+carries updated priority parameters and is sent on the version-specific control
+stream. The frame references the response to reprioritize based on a
+version-specific identifier; in HTTP/2 this is the Stream ID, in HTTP/3 this is
+either the Stream ID or Push ID.
+
+## HTTP/2 REPRIORITY Frame
+
+The HTTP/2 REPRIORITY frame (type=0xF) carries the stream ID of the response that
+is being reprioritized, and the updated priority in ASCII text, using the same
+representation as that of the Priority header field value.
+
+The REPRIORITY frame is sent on stream 0.
+
+~~~ drawing
+  0                   1                   2                   3
+  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+ +---------------------------------------------------------------+
+ |R|                        Stream ID (31)                       |
+ +---------------------------------------------------------------+
+ |                   Priority Field Value (*)                  ...
+ +---------------------------------------------------------------+
+~~~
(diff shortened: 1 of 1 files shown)


kazuho (approved):
Thank you for the PR. Looks good to me.

draft-kazuho-httpbis-priority.md
```diff
@@ -351,15 +353,31 @@ render textual information at an early moment.
...
 
-Therefore, to support reprioritization, it is necessary to define a
+However, a client cannot reprioritize a response by using the Priority header
+field.  This is because an HTTP header field can only be sent as part of an HTTP
+message. Therefore, to support reprioritization, it is necessary to define a
 HTTP-version-dependent mechanism for transmitting the priority parameters.
 
-One approach that we can use in HTTP/2 ({{?RFC7540}}) is to use a frame that
-carries the priority parameters.
+One approach that we can use in HTTP/2 ({{?RFC7540}}) and HTTP/3
```
kazuho: Instead of suggesting it being "one approach", maybe state that the frame-based design *is* the approach we have adopted?

kazuho (commented):
Rereading, I think you might want to move the Reprioritization section up, outside of Considerations, and suggest that this is the design the draft has adopted.

draft-kazuho-httpbis-priority.md
LPardue: Agreed

rmarx:
Hey Lucas, 
Thanks for putting this together. Some comments: 

1. I'm not too sure about the name REPRIORITY, it feels a bit weird. Something like REPRIORITIZE sounds better, but I get that isn't consistent with the PRIORITY frame name in the first place (though GOAWAY and CONNECT are also more verb-y...). PRIORITY_CHANGE maybe? 

2. I find the 7-bit empty space a bit weird... in a separate discussion you said it's because implementation-wise, you can't read from the transport bit-by-bit, only per-byte, and this makes that easy. In that case, I feel the rationale should be at least mentioned here. However, even then, I wonder if 2 separate frame types doesn't make more sense for H3 (similar to CONNECTION_CLOSE in QUIC for example). 

3. I feel the reprioritization example of the HTML file a bit unrealistic maybe. Are people actually prefetch-ing whole documents? I personally would change it to a prefetch-ed JavaScript request or maybe an image that was off-screen but now the user started scrolling down.  

4. I'm probably not enough into QPACK for this, but: why is the updated priority header here plaintext ASCII instead of a compressed header block? Again, if there is a good reason, I would prefer that the text reference that reason.

LPardue:
> I'm not too sure about the name REPRIORITY, it feels a bit weird. Something like REPRIORITIZE sounds better, but I get that isn't consistent with the PRIORITY frame name in the first place (though GOAWAY and CONNECT are also more verb-y...). PRIORITY_CHANGE maybe?

The more I think about, the more I dislike REPRIORITY. I'm leaning towards PRIORITY_UPDATE. What do people think? We should also bear in mind that we are aiming to rename the document, so this frame name should probably reflect that too.

> I find the 7-bit empty space a bit weird... in a separate discussion you said it's because implementation-wise, you can't read from the transport bit-by-bit, only per-byte, and this makes that easy. In that case, I feel the rationale should be at least mentioned here. However, even then, I wonder if 2 separate frame types doesn't make more sense for H3 (similar to CONNECTION_CLOSE in QUIC for example).

Actually, the first design I came up with emulated the CONNECTION_CLOSE frame, which effectively reserves two codepoints and uses that to determine the applicability of the frame. I chose not to go with that design because I wanted a simple compatibility between H2 and H3. I agree the 7 bits are pretty wasted but is that such a problem? 

> I feel the reprioritization example of the HTML file a bit unrealistic maybe. Are people actually prefetch-ing whole documents? I personally would change it to a prefetch-ed JavaScript request or maybe an image that was off-screen but now the user started scrolling down.

I opened #53 to track this seperately. Because the text was like that before anyway.

> I'm probably not enough into QPACK for this, but: why is the updated priority header here plaintext ASCII instead of a compressed header block? Again, if there is a good reason, I would prefer that the text reference that reason.

I think @kazuho and I have discussed this a bit previously. While we could have such a design my gut feeling is that it adds complexity for small gain. A good example is the ALTSVC frame, which also uses the ASCII version. Specific to H3, by making this QPACK encoded we introduce more coupling between the control streams and QPACK streams; right now we don't really have that much dependency between these streams. I then might also have a knock-on effect from control stream to request stream. For these reason I'd be more enthusiastic about reducing the verbosity of the Priority Header parameters.

kazuho:
> The more I think about, the more I dislike REPRIORITY. I'm leaning towards PRIORITY_UPDATE. What do people think? We should also bear in mind that we are aiming to rename the document, so this frame name should probably reflect that too.

I do not have a strong opinion, but "PRIORITY" + another word does sound better, as it better reflects the fact that it is a change to values sent in the "Priority" header field.

> Actually, the first design I came up with emulated the CONNECTION_CLOSE frame, which effectively reserves two codepoints and uses that to determine the applicability of the frame. I chose not to go with that design because I wanted a simple compatibility between H2 and H3. I agree the 7 bits are pretty wasted but is that such a problem?

Up to https://tools.ietf.org/html/draft-ietf-quic-http-22, we had empty bits in the PRIORITY frame. Therefore I think it would be less surprising to use the empty bits rather than using two frame types.

Re compressing the field value transmitted using the REPRIORITY frame, I think we might want to merge this PR as-is then open a separate issue.

LPardue:
I pushed https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/51/commits/ac05059139952df94620a5b65b357f452d41cb0f that renames the frame from REPRIORITY to PRIORITY_UPDATE. I think we are good to merge, PTAL.

> Re compressing the field value transmitted using the REPRIORITY frame, I think we might want to merge this PR as-is then open a separate issue.

Agreed, we won't don't have the time to resolve that discussion, I'll open a ticket to track it

draft-kazuho-httpbis-priority.md
```diff
@@ -300,6 +302,84 @@ receipt, a client that supports header-based prioritization MUST close the
 connection with a protocol error. Non-supporting clients will ignore this
 extension element (see {{!RFC7540}}, Section 5.5).
 
+# Reprioritization
+
+Once a client sends a request with a priority, circumstances might change and
```
kazuho: > Once a client sends a request with a priority

I think we might want to strike "with a priority," as even a request without the Priority header field carries an implicit priority.

draft-kazuho-httpbis-priority.md
```diff
@@ -300,6 +302,84 @@ receipt, a client that supports header-based prioritization MUST close the
...
+Then, when the user navigates to the HTML while prefetch is in action, it would
+send a reprioritization frame with the priority field value set to `urgency=0`.
+
+However, a client cannot reprioritize a response by using the Priority header
+field.  This is because an HTTP header field can only be sent as part of an HTTP
+message. Therefore, to support reprioritization, it is necessary to define a
+HTTP-version-dependent mechanism for transmitting the priority parameters.
+
+This document specifies a new frame type for HTTP/2 ({{?RFC7540}}) and HTTP/3
+({{?I-D.draft-ietf-quic-http-23}}) that is specialized for reprioritization. It
```
kazuho: Maybe we need to refer to them in a normative fashion (`!`) as we are defining extensions of the two?

draft-kazuho-httpbis-priority.md
```diff
@@ -300,6 +302,84 @@ receipt, a client that supports header-based prioritization MUST close the
...
+carries updated priority parameters and is sent on the version-specific control
+stream. The frame references the response to reprioritize based on a
```
kazuho: I think we might want to delete "and is sent on the version-specific control stream."

HTTP/2 does not have a control stream. Stream ID 0 is an identifier indicating that the frame does not belong to *any* stream.

rmarx:
LGTM

draft-kazuho-httpbis-priority.md
```diff
@@ -300,6 +302,84 @@ receipt, a client that supports header-based prioritization MUST close the
...
+version-specific identifier; in HTTP/2 this is the Stream ID, in HTTP/3 this is
+either the Stream ID or Push ID.
+
+## HTTP/2 PRIORITY_UPDATE Frame
+
+The HTTP/2 PRIORITY_UPDATE frame (type=0xF) carries the stream ID of the
+response that is being reprioritized, and the updated priority in ASCII text,
+using the same representation as that of the Priority header field value.
+
+The PRIORITY_UPDATE frame is sent on stream 0.
```
kazuho: "The stream identifier for a PRIORITY_UPDATE frame MUST be zero (0x0)."

kazuho (commented):
Thank you for the updates. LGTM modulo my comments below.

draft-kazuho-httpbis-priority.md
LPardue: excellent point! fixed

draft-kazuho-httpbis-priority.md
LPardue: good point I agree. I'd like to highlight that this frame is, unlike the PRIORITY, not allowed on request streams. I'll take a stab at changing the wording around in a separate commit but if I can't get it right I will delete the offending statement you have highlighted. I have a few TODOs in the document to flesh out frame and stream handling requirements and we could defer this to them.

draft-kazuho-httpbis-priority.md
LPardue: fixed, and tidied up all references to QUIC and HTTP/2 docs

draft-kazuho-httpbis-priority.md
LPardue: fixed

draft-kazuho-httpbis-priority.md
LPardue: Had an attempt on https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/51/commits/f5c7635060742e74f4d814ebd122ca53d76faee1

draft-kazuho-httpbis-priority.md
LPardue: I realised this might be a little confusing because the H2 frame header includes the Stream Identifier field, and the frame payload contains a Stream ID field. So in https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/51/commits/f5c7635060742e74f4d814ebd122ca53d76faee1 I made it explicit.

kazuho:
Thank you for the updates. Looks good to me. I'm going to merge this, hoping that reducing the number of open PRs would help us move forward.

PR #52, opened 2019-10-30
clarify why indicating the use of header-based prioritization is beneficial

kazuho:
Aside from if this signal should be carried as part of draft-lassey-priority-setting, I think it's worth clarifying why it is a good idea to have a SETTINGS Parameter.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -276,21 +276,27 @@ prioritization as specified in this document.
 
 ## The SETTINGS_HEADER_BASED_PRIORITY SETTINGS Parameter
 
-To improve communication of the client's intended prioritization scheme, this
-document specifies a new HTTP/2 SETTINGS parameter with the name
-`SETTINGS_HEADER_BASED_PRIORITY`. The value of the parameter MUST be 0 or 1; the
-initial value is 0. Frame-based prioritization is respected when the value is 0,
-or when the server does not recognize the setting.
+This header-based prioritization scheme is the second prioritization scheme that
+can be applied to HTTP/2. The fact introduces a complexity to the server
+implementations unless the server can determine which of the two priorization
+scheme is being used by the client before receiving the first request on an
+HTTP/2 connection, as the server would be required to prepare for both schemes
+and possibly balance the responses between the two groups.
 
-An HTTP/2 client that uses header-based priority SHOULD send a
-`SETTINGS_HEADER_BASED_PRIORITY` parameter with a value of 1 when connecting to
-a server.
+To address the concern, this document specifies a new HTTP/2 SETTINGS parameter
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -276,21 +276,27 @@ prioritization as specified in this document.
...
-value of 1 for a connection it establishes when, and only when, all the requests
-to be sent over that connection originate from a client that utilizes this
-header-based prioritization scheme. Otherwise this settings parameter SHOULD be
-set to 0.
+This header-based prioritization scheme is the second prioritization scheme that
+can be applied to HTTP/2. The fact introduces a complexity to the server
+implementations unless the server can determine which of the two priorization
+scheme is being used by the client before receiving the first request on an
+HTTP/2 connection, as the server would be required to prepare for both schemes
+and possibly balance the responses between the two groups.
```
LPardue, line 284: Reading between the lines, I think what you are saying is that we expect there to be only one prioritization scheme in effect per connection regardless of how many signals are sent by the client. Therefore, should we state that explicitly, perhaps after the first sentence?

draft-kazuho-httpbis-priority.md
```diff
@@ -276,21 +276,27 @@ prioritization as specified in this document.
...
+
+To address the concern, this document specifies a new HTTP/2 SETTINGS parameter
+called `SETTINGS_HEADER_BASED_PRIORITY`. The value of the parameter MUST be 0 or
+1; the initial value is 0. A value of one indicates he HTTP/2 client's intent to
```
LPardue, line 288:
```suggestion
1; the initial value is 0. A value of 1 indicates the HTTP/2 client's intent to
```

consistently use `1` and fix typo

draft-kazuho-httpbis-priority.md
```diff
@@ -276,21 +276,27 @@ prioritization as specified in this document.
...
+use the header-based prioritization scheme defined in this document. A value of
+zero indicates the other.
+
+When the client indicates its intent to use this header-based prioritization
+scheme, it SHOULD ignore the SETTINGS frame sent by the client.
```
LPardue, line 293: I think there's a typo in here because it reads like the client ignores the client, which seems weird.

LPardue (changes requested):
Some nits identified in the text. Agreed we can decide where the text belongs later.

draft-kazuho-httpbis-priority.md
kazuho: Thank you for stating that. I think that is the good way of summarizing the paragraph. I’ll try to accommodate that nuance.

kazuho:
Following our discussion, I assume that the points will be covered by draft-lassey-priority-setting.

@LPardue I'm keeping this PR open just in case you want to use this PR as a remainder to have the points reflected in draft-lassey-priority-setting. Please close if that's unnecessary.

kazuho:
Subsumed by #58.

PR #54, opened 2019-10-31
change title and abstract

kazuho:
The title is changed to the one suggested by @ianswett. The PR also updates the abstract reflecting the change of the title.

@ianswett @LPardue PTAL.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -29,7 +29,7 @@ informative:
 
 --- abstract
 
-This document describes a scheme for prioritizing the delivery of the HTTP
+This document describes a scheme for prioritizing the delivery of HTTP
 responses. The prioritization scheme expresses absolute values for each HTTP
 response, rather than the relative precedence between a group of HTTP responses.
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -1,5 +1,5 @@
 ---
-title: Extensible Prioritization for HTTP
+title: Extensible Prioritization Scheme for HTTP
 docname: draft-kazuho-httpbis-priority-latest
 category: std
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -31,8 +31,7 @@ informative:
 
 This document describes a scheme for prioritizing the delivery of the HTTP
 responses. The prioritization scheme expresses absolute values for each HTTP
-response, rather than the relative relationship between a group of HTTP
-responses.
+response, rather than the relative precedence between a group of HTTP responses.
 
 This document defines the Priority header field for communicating the initial
 priority in an HTTP version-independent manner, as well as HTTP/2 and HTTP/3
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -29,9 +29,15 @@ informative:
...
-in an HTTP-version-independent way.
+This document describes a prioritization scheme for the HTTP protocol that is
+independent to the HTTP version being used.  It defines a header field used to
+communicate the initial priority, as well as HTTP/2 and HTTP/3 frames for
+reprioritizing the responses.
+
+The priority is expressed using absolute values for each HTTP message, rather
+than expressing the relative precedence between a group of HTTP messages.  The
+structure of the Priority header field and the frames is designed to provide
+future extensibility.
```
LPardue: I like this text but I might change the order of words a little bit. Just my 2c, perhaps too closed-ended :

"This document describes a version-independent prioritization scheme for the HTTP protocol that applies to the resource allocation of HTTP response message payload body. The scheme expresses absolute values for each HTTP message, rather than expressing the relative precedence between a group of HTTP messages. 

This document defines the Priority header field used to communicate the initial priority, as well as HTTP/2 and HTTP/3 frames for reprioritization. These share a common format structure that is designed to provide future extensibility."

LPardue: I was purposefully vague on whether the response resource allocation affects server or client, because we might want to inform client behaviour when it comes to reading.

kazuho: Thank you for the suggestion. I like the shuffling, and updated the abstract in commits up to 5fae1f1.

* I've gone with "delivery" rather than "resource allocation", as the latter seemed a bit confusing.
* s/HTTP response message payload body/HTTP response/. I do not think we (need to) specify if the prioritization scheme is specific to the body (but not to the headers).
* Moved "version independent" to the description of the header field. Frames are definitely version-dependent.

draft-kazuho-httpbis-priority.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: The Priority HTTP Header Field
+title: Extensible Prioritization for HTTP
```
kazuho: Should this rather be "Extensible Prioritization Scheme for HTTP"?

LPardue: I think that sounds more true to the nature of the document

draft-kazuho-httpbis-priority.md
```diff
@@ -29,9 +29,14 @@ informative:
 
 --- abstract
 
-This document describes the Priority HTTP header field.  This header field can
-be used by endpoints to specify the absolute precedence of an HTTP response
-in an HTTP-version-independent way.
+This document describes a scheme for prioritizing the delivery of the HTTP
```
```suggestion
This document describes a scheme for prioritizing the delivery of HTTP
``` 

superfluous "the" maybe?

LPardue (approved):

draft-kazuho-httpbis-priority.md
kazuho: Thanks! Applied in aef4574.

LPardue:
:shipit:

kazuho:
🚢 ed

ianswett:
Thanks!

PR #55, opened 2019-10-31
negotiation is defined in priority-settings

kazuho:
This PR
* removes the definition of the SETTINGS Parameter used to indicate the use of this prioritization scheme
* instead references https://datatracker.ietf.org/doc/draft-lassey-priority-setting/
  * states that ID 0xTBD is used as the identifier for this prioritization scheme
* recommends endpoints to send the Priority header even if the per-hop prioritization scheme is going to be something different

Regarding negotiation, one of the things that this PR removes is a statement that says that a coalescing intermediary should forward the negotiation signal from the client only if all the clients send a common signal: _An intermediary SHOULD send a `SETTINGS_HEADER_BASED_PRIORITY` parameter with a value of 1 for a connection it establishes when, and only when, all the requests
to be sent over that connection originate from a client that utilizes this header-based prioritization scheme. Otherwise this settings parameter SHOULD be set to 0._ I presume that the Priority Setting I-D would have a similar statement.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -188,7 +188,7 @@ An 8 bit value of 1 in HTTP/2 indicates support for HTTP/2 priorities
 as defined in Section 5.3 of [RFC7540] and is an error in HTTP/3 because
 there is not a clean mapping to HTTP/3.
 
-## Negotiating the Extensible Priority Scheme
+## Negotiating the Extensible Priority Scheme {#settings-this-scheme}
 
 The extensible priority scheme is negotiated using the described mechanism. It is
 identified by the 8-bit value of 2.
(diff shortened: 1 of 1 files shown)


LPardue (approved):
In general I think this looks great. However, I'd be inclined to hold off merging the PR until we have some time to review the update to negotiation draft.

kazuho:
@LPardue I think this PR is ready for review.

Please note that we no longer say something like, quoting from the original text: _an intermediary SHOULD send a `SETTINGS_HEADER_BASED_PRIORITY` parameter with a value of 1 for a connection it establishes when, and only when, all the requests to be sent over that connection originate from a client that utilizes this header-based prioritization scheme._

Instead, we say that should be done _when and only when all the requests that are to be sent on that backend connection originates from one client-side connection that has negotiated the use of the extensible priority scheme_.

The latter is stricter, and the rationale is being described in the fairness section (within Security Considerations).

LPardue:
So just to confirm my understanding, when we say 

> when and only when all the requests that are to be sent on that backend connection originates from one client-side connection that has negotiated the use of the extensible priority scheme.

this means a intermediary that coalesces client connections does not add the 8-bit value 2 to their SETTINGS_PRIORITIES frame?

kazuho:
@LPardue That's correct. I think I've argued that in some other PR, added clarification in 452890f.

Though I can now see that it might be confusing, so I do not mind taking a step back and using SETTINGS to communicate if all clients are using the header-based scheme.

LPardue:
Let's put it in as you've written - its an important design consideration and it would be good to solicit feedback on this behaviour, even if that means we back it out or modify it later.

LPardue (approved):

kazuho:
Thank you!!!

PR #56, opened 2019-10-31
explain the fairness vs. coalescing proxy issue and the mitigations

kazuho:
I still need to go through the notes to collect the issues we've discussed regarding the use of header fields, but I think this PR explains the biggest issue as well as the mitigations.

PTAL.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -316,12 +316,12 @@ one HTTP/2 or HTTP/3 connection going to the backend server, requests that
 originate from one client might have higher precedence than those coming from
 others.
 
-It is sometimes beneficial for the server running behind an intermediary to obey to the value of
-the Priority header field. As an example, a resource-constrained server might
-defer the transmission of software update files that would have the background
-urgency being associated. However, in the worst case, the asymmetry between the
-precedences declared by multiple clients might cause responses going to one end
-client to be delayed totally after those going to another.
+It is sometimes beneficial for the server running behind an intermediary to obey
+to the value of the Priority header field. As an example, a resource-constrained
+server might defer the transmission of software update files that would have the
+background urgency being associated. However, in the worst case, the asymmetry
+between the precedences declared by multiple clients might cause responses going
+to one end client to be delayed totally after those going to another.
 
 In order to mitigate this fairness problem, a response to a request that is
 known to have come through an intermediary SHOULD be assigned the priority of
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -333,6 +333,7 @@ A server can determine if a request came from an intermediary through
 configuration, or by consulting if that request contains one of the following
 header fields:
 
+* CDN-Loop ({{?RFC8586}})
 * Forwarded, X-Forwarded-For ({{?RFC7239}})
 * Via ({{?RFC7230}}, Section 5.7.1)
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -323,11 +323,12 @@ background urgency being associated. However, in the worst case, the asymmetry
 between the precedences declared by multiple clients might cause responses going
 to one end client to be delayed totally after those going to another.
 
-In order to mitigate this fairness problem, a response to a request that is
-known to have come through an intermediary SHOULD be assigned the priority of
-`urgency=0, progressive=?1` (i.e. round-robin), unless the server has the
-knowledge that the intermediary is not coalescing requests from multiple
-clients.
+In order to mitigate this fairness problem, when a server responds to a request
+that is known to have come through an intermediary, the server SHOULD prioritize
+the response as if it was assigned the priority of  `urgency=0, progressive=?1`
+(i.e. round-robin) regardless of the value of the Priority header field being
+transmitted, unless the server has the knowledge that no intermediaries are
+coalescing requests from multiple clients.
 
 A server can determine if a request came from an intermediary through
 configuration, or by consulting if that request contains one of the following
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -314,7 +314,7 @@ the Priority header field. As an example, a resource-constrained server might
 defer the transmission of software update files that would have the background
 urgency being associated. However, in the worst case, the asymmetry between the
 precedences declared by multiple clients might cause responses going to one end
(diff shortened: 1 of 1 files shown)


rmarx:
I like the recommendation of doing round robin to provide fairness, the mention of weighted round robin for smarter servers and the mention of the bottleneck location. 

I'm not sure if this fully alleviates the whole end-to-end vs hop-by-hop discussions we've had though (or was that not the intent?). I'm also not 100% sure how this fits with server-sent priorities in the case where you have more than 1 intermediary... or maybe we should just forget about that super-hypothetical? 

The main remark I have though is that the buildup of the argument in the text seems a bit off. You first mention that requests from certain clients can have higher precedence and give examples and that it can be beneficial for the server to follow the original Priority values. Then, you seem to undo all of that by saying that all coalesced requests SHOULD be round-robin'ed, regardless of whether they have a valid higher precedence or not. Then you back out of that a bit by telling they MAY do it weighted if considered useful. As an implementer, I would find this section quite confusing as to what to do with this in practice.

kazuho:
@rmarx Thank you for the feedback.

> I'm not sure if this fully alleviates the whole end-to-end vs hop-by-hop discussions we've had though (or was that not the intent?).

No, this is still incomplete.

> I'm also not 100% sure how this fits with server-sent priorities in the case where you have more than 1 intermediary... or maybe we should just forget about that super-hypothetical?

I do not think handling of server-sent priorities is going to have any affect on the rules here. Prioritization is about sending responses. It's just either obeying to the value in request header or the response header.

> The main remark I have though is that the buildup of the argument in the text seems a bit off.

Yeah. Let me see if I can refactor as I add other topics.

Though my excuse might be that the discussion about fairness is something that should go into Security Considerations. And in a Security Considerations, it is important to explain what the concern is.

draft-kazuho-httpbis-priority.md
```diff
@@ -300,6 +300,44 @@ receipt, a client that supports header-based prioritization MUST close the
...
+`urgency=0, progressive=?1` (i.e. round-robin), unless the server has the
+knowledge that the intermediary is not coalescing requests from multiple
+clients.
+
+A server can determine if a request came from an intermediary through
+configuration, or by consulting if that request contains one of the following
+header fields:
+
+* Forwarded, X-Forwarded-For ({{?RFC7239}})
+* Via ({{?RFC7230}}, Section 5.7.1)
```
LPardue, line 339: CDN loop?

LPardue:
> I'm also not 100% sure how this fits with server-sent priorities in the case where you have more than 1 intermediary... or maybe we should just forget about that super-hypothetical?

Chains of intermediaries was the only comment I could make on this PR. Specifically, 

> response to a request that is known to have come through an intermediary SHOULD be assigned the priority of `urgency=0, progressive=?1` (i.e. round-robin),

What does it mean to assign a priority? Does that mean it overwrites the value of the Priority header in request that the intermediary passes to the upstream? Or is it more like "always pass on the input Priority, but intermediaries SHOULD pass back response data as round robin, and edge servers SHOULD pass back response data as indicated by the final final combination of Priority header (client request header modulated by intermediary response header)"?

rmarx:
So, thinking about this a bit more... how do you know the intermediary is coalescing or not? Is there a simple way to do this? If we assume the intermediary terminates the original client QUIC connections and maps client-to-intermediary streams to new intermediary-to-server streams, I can't really see how detection could be done without an explicit signal. 

As such, the advise you're giving now, is maybe the wrong way around. Now it's: You should round-robin, unless you know it's not coalesced (which is hard).

Maybe it should be: You should respect the original priorities, unless you know it IS coalesced (for which we might want to define an explicit signal), then you should round-robin.

Overall, I feel this flits too much into the end-to-end vs hop-by-hop debate and e.g., @MikeBishop's proposal of double headers (e.g., X-original-priority) would help solve this problem.

rmarx:
See also @MikeBishop's new issue at #57.

kazuho:
@LPardue 
> What does it mean to assign a priority? Does that mean it overwrites the value of the Priority header in request that the intermediary passes to the upstream? Or is it more like "always pass on the input Priority, but intermediaries SHOULD pass back response data as round robin, and edge servers SHOULD pass back response data as indicated by the final final combination of Priority header (client request header modulated by intermediary response header)"?

Good point. I think the intent is the latter, based on the principle that the priority header is the intent of the originator of the request. I'll update the text accordingly.

@rmarx 
> Maybe it should be: You should respect the original priorities, unless you know it IS coalesced (for which we might want to define an explicit signal), then you should round-robin.

That does not address the concern, as there could be a coalescing intermediary that does not implement this specification.

To be safe, what we need to say is "do something unless you know that it IS NOT coalesced."

That's why I think this PR is the baseline.

We could optimize for cases where some or all the intermediaries implement this specification. #57 is good in that sense. At the moment, we also state that SETTINGS_HEADER_BASED_PRIORITY can be used to signal that all the intermediaries are not coalescing, though we need to determine if / how we want to retain the property in the new negotiation scheme. But they are all optimizations IMO.

draft-kazuho-httpbis-priority.md
kazuho: Good point! Done in c8dbd04.

kazuho:
@MikeBishop @rmarx Based on my observation that #57 is an optimization of the Security Considerations written in this PR, I've added what I believe is the proposal of #57 to Considerations.

WDYT?

draft-kazuho-httpbis-priority.md
```diff
@@ -307,6 +307,45 @@ receipt, a client that supports header-based prioritization MUST close the
...
+# Security Considerations
+
+## Fairness and Coalescing Intermediaries {#fairness}
+
+When an intermediary coalesces HTTP requests coming from multiple clients into
+one HTTP/2 or HTTP/3 connection going to the backend server, requests that
+originate from one client might have higher precedence than those coming from
+others.
+
+It is sometimes beneficial for the server running behind to obey to the value of
```
```suggestion
It is sometimes beneficial for the server running behind an intermediary to obey to the value of
```

Just a little suggestion for readability. I know this is in a section that discusses intermediaries but on two separate readings the sentence construct made me pause to think

LPardue (approved):
lgtm modulo the one suggested improvement

kazuho:
@LPardue Thank you for the suggestions. 8193f8d should address your concern raised in https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/56#issuecomment-548834562. Would you mind checking?

LPardue:
Yeah that is a big helpful clarification (to me anyway). Thanks.

Issue #50, opened 2019-09-30
Signal to indicate preferred prioritization scheme


rmarx:
Since it looks like this draft will grow into the full new H3 proposal, I am posting this somewhat more tangential issue here as well. Let's dub it the "scheme selection signal". 

One of the things we keep seeing in our synthetic prioritization tests, is that the optimal scheme (scheme = round-robin, fifo, a specific combination of the two, see https://h3.edm.uhasselt.be/images/streams.png) can be **different per web page**. This is a problem, since for H2, each browser basically hard-codes a single scheme and servers are expected to follow it. This means some pages load much worse in some browser than others. Developers should thus ideally be able to tweak the browser (or server) behavior for certain use cases. 

The most practical solution we came up with in our [paper](https://h3.edm.uhasselt.be/) (since creating full prioritized lists of all resources, either manually or automatically, isn't very feasible), is to allow developers to select one of x different "schemes", indicating that scheme performs best for a particular page (e.g., for this page, you should use a weighted round-robin, for that one Patrick Meenan's structure). The optimal scheme can be determined practically by e.g., loading the pages offline using something like Webpagetest, and seeing if something sticks out positively (or negatively). 

As such, you would need a **second signal** (next to #49), indicating not how/which signals you wish to receive (e.g., dependency tree vs HTTP headers from this document), but rather **how to act on those signals.** For example, this document's approach allows servers much more leeway than H2's tree in how to actually interpret the prioritization signals. An urgency of 'prerequisite' could be interpreted as "send in full before the rest" but also as "add to Round-Robin with highest possible weight".  

To make this work, I envision that next to the "scheme selection signal", we would also have to define **a list of "default" schemes**, identifiable by a string name, and their globally expected behavior. I feel we might want such a list or at least some examples **either way** (even if we don't do this second signal), to guide implementers on how to actually implement the new approach (how can/should a server translate 'urgency' and 'progressive' into bandwidth distribution). I fear that otherwise, in combination with the very unspecified quic-layer prioritization, this work will have sub-optimal impact, as implementers will choose the most direct/naive mapping of the signals to bandwidth-level behavior. 

[...]

Issue #28, opened 2019-07-09
Header fields and values to support


rmarx:
The current draft (-00) only supports a limited set of header fields and values (urgency, with 3 possible values and progressive, a boolean). Several comments have indicated that urgency should probably need at least 2 extra values (deferred and background).

I take some issue with this, because the current discourse seems to be mostly "we need 5 because Chrome uses 5" and I see some comments from @kazuho that discuss how these values can be mapped to bandwidth distributions (and, in my interpretation then, because bandwidth can only be distributed a limited number of ways, we don't need many header fields/values).

The way I see it, this part of the proposal should be general-purpose, communicating plenty of resource metadata to the server so more complex/fine-grained setups are possible if the server so desires. It can be debated how plentiful this should be, but imo, more flexible than what we have now. I feel that, in deciding which metadata to send and at which granularity, we should not just be looking at how this can concretely impact the bandwidth distribution (or how this maps to current browser approaches) but from a higher-level. 
___

A strawman proposal for this would be to map html-parser level semantics (currently used by browsers to build their heuristics) semi-directly onto the headers. Some of this directly reflects/echos back metadata set in the HTML (e.g., whether something comes from a preload <link> element or whether a JS file had the "async" attribute), while some come from internal browser implementation specifics (e.g., parser/render-blocking, whether a resource can be processed/parsed incrementally, ...). 

This would lead to (semi-verbose) lists of **"aspects"** to be set for each resource.
For example, if we assume the defaults of: 
`aspects=non-blocking,non-incremental,immediate,body`

- For a CSS file in the <head>: `aspects=render-blocking,head`
- For an image file above the fold: `aspects=incremental,visible` (below the fold would loose 'visible')
- For a pre-loaded Font in the <head>: `aspects=preload,visible` (preload implies 'delayed' vs the default 'immediate')
- For a defer-ed JS on the bottom of the HTML: `aspects=defer,non-immediate`  (a JS tagged as async would be `aspects=async`)

The nice thing about this approach is that it prevents us from having to re-define anything semantically: the mapping from inherent browser semantics, known by most developers (see #27), is kept intact. This makes the header values easy to follow/understand/manipulate for both browser implementers and end-users/web developers, leaving less room for errors/footguns. 

The main downsides I can see are:
- verbosity: although that seems quite manageable in the examples above with sensible defaults + QPACK might help there in practice
[...]

PR #58, opened 2019-11-01
Negotiating priorities using a new SETTING

ianswett:
Summarizes what I understood from the last design team meeting.  Uses 8 bit codepoints to specify priority schemes in a new HTTP/2 and HTTP/3 SETTING.

I did make a design change so the server's preference to win over the client's, based on the belief that the server is the consumer of the information the client is sending, so knows what information is most valuable.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -143,8 +143,8 @@ lack thereof of priority frames.
 If the value is non-zero, then the least significant 8 bits indicates the
 peer's preferred priority scheme, the second least significant 8 bits
 indicates the peer's second choice, and so on.  This allows expressing
-support for 4 schemes in HTTP/2 and 7 in HTTP/3.  If the value is non-zero,
-all 8 bit scheme identifiers MUST NOT be 0.
+support for 4 schemes in HTTP/2 and 7 in HTTP/3.  If any octet is 0,
+all more significant octets MUST also be 0.
 
 In HTTP/2, the setting SHOULD appear in the first SETTINGS frame and peers
 MUST NOT process the setting if it's received multiple times in order to
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -105,22 +105,26 @@ Section 5.3, which some implementers have opted not to fully support.
 The lack of signalling about the status of the implementation has
 caused several implementations to implement heuristics to detect when
 the clients they are connected to do not support priorities as
-defined and take steps to compensate for that.  The intent of this
-negotiation is to provide an affirmative signalling mechanism for each
-peer to communicate which, if any, priority schemes are supported,
-including the priority scheme as defined in [RFC7540], Section 5.3.
+defined and take steps to compensate for that.
+
+The intent of this negotiation is to provide a signalling mechanism
+for each peer to communicate which, if any, priority schemes are
+supported, as well as a ranked preference.
 
 For both HTTP/2 and HTTP/3, either peer's SETTINGS may arrive first,
 so any negotiation must be unilateral and not rely upon receiving
 the peer's SETTINGS value.
 
 Implementations are likely to only use one prioritization scheme at
-once, so the SETTINGS SHOULD be sent prior to the first request.
-In HTTP/3, the SETTINGS may arrive after the first request even if
-they are sent first.  In order to avoid the server choosing
-incorrectly in HTTP/3, the client SHOULD only send priority
-information via its most preferred scheme until it knows what is
-supported by the server.
+once, and may be unable to change the scheme once established, so the
+setting MUST be sent prior to the first request if it is ever sent.
+In HTTP/3, SETTINGS may arrive after the first request even if
+they are sent first.  In order to avoid the server incorrectly
+choosing a priority scheme in HTTP/3, the client SHOULD only send
+priority information via its most preferred scheme until it knows
+what is supported by the server.  The client may also delay
+sending priority information until after it knows what the server
+supports.
 
 ## The SETTINGS_PRIORITIES SETTINGS Parameter
 
@@ -138,7 +142,8 @@ lack thereof of priority frames.
 If the value is non-zero, then the least significant 8 bits indicates the
 peer's preferred priority scheme, the second least significant 8 bits
 indicates the peer's second choice, and so on.  This allows expressing
-support for 4 schemes in HTTP/2 and 7 in HTTP/3.
+support for 4 schemes in HTTP/2 and 7 in HTTP/3.  If the value is non-zero,
+all 8 bit scheme identifiers MUST NOT be 0.
 
 In HTTP/2, the setting SHOULD appear in the first SETTINGS frame and peers
 MUST NOT process the setting if it's received multiple times in order to
@@ -148,9 +153,9 @@ If there is a prioritization scheme supported by both the client and server,
 then the client's preference order prevails and both peers SHOULD
 only use the agreed upon priority scheme for the remainder of the session.
 
-An 8 bit value of 1 in HTTP/2 indicates support for HTTP/2 priorities as defined
-in Section 5.3 of [RFC7540] and is an error in HTTP/3 because there is not
-a clean mapping to HTTP/3.
+An 8 bit value of 1 in HTTP/2 indicates support for HTTP/2 priorities
+as defined in Section 5.3 of [RFC7540] and is an error in HTTP/3 because
+there is not a clean mapping to HTTP/3.
 
 
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -153,7 +153,7 @@ avoid changing the agreed upon prioritization scheme.
 If there is a prioritization scheme supported by both the client and server,
 then the servers's preference order prevails and both peers SHOULD
 only use the agreed upon priority scheme for the remainder of the session.
-The server chooses becasue it is in the best position to know what
+The server chooses because it is in the best position to know what
 information from the client is of the most value.
 
 An 8 bit value of 1 in HTTP/2 indicates support for HTTP/2 priorities
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -109,7 +109,7 @@ defined and take steps to compensate for that.
 
 The intent of this negotiation is to provide a signalling mechanism
 for each peer to communicate which, if any, priority schemes are
-supported, as well as a ranked preference.
+supported, as well as the client's ranked preference.
 
 For both HTTP/2 and HTTP/3, either peer's SETTINGS may arrive first,
 so any negotiation must be unilateral and not rely upon receiving
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -98,6 +98,62 @@ The terms sh-token and sh-boolean are imported from
 Example HTTP requests and responses use the HTTP/2-style formatting from
 {{?RFC7540}}.
 
+# Negotiating Priorities
+
+The HTTP/2 specification defines a priority scheme in [RFC7540],
+Section 5.3, which some implementers have opted not to fully support.
+The lack of signalling about the status of the implementation has
+caused several implementations to implement heuristics to detect when
+the clients they are connected to do not support priorities as
+defined and take steps to compensate for that.  The intent of this
+negotiation is to provide an affirmative signalling mechanism for each
+peer to communicate which, if any, priority schemes are supported,
+including the priority scheme as defined in [RFC7540], Section 5.3.
+
+For both HTTP/2 and HTTP/3, either peer's SETTINGS may arrive first,
+so any negotiation must be unilateral and not rely upon receiving
+the peer's SETTINGS value.
+
+Implementations are likely to only use one prioritization scheme at
+once, so the SETTINGS SHOULD be sent prior to the first request.
+In HTTP/3, the SETTINGS may arrive after the first request even if
+they are sent first.  In order to avoid the server choosing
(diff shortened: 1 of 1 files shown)


LPardue:
For watchers, IIUC This PR incorporates all negotiation consideration into kazuhos draft. The purpose being to provide a single document for wider dissemination and discussion. At a later stage we can decide our preferred document structuring.

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,62 @@ The terms sh-token and sh-boolean are imported from
...
+# Negotiating Priorities
+
+The HTTP/2 specification defines a priority scheme in [RFC7540],
+Section 5.3, which some implementers have opted not to fully support.
+The lack of signalling about the status of the implementation has
+caused several implementations to implement heuristics to detect when
+the clients they are connected to do not support priorities as
+defined and take steps to compensate for that.  The intent of this
+negotiation is to provide an affirmative signalling mechanism for each
+peer to communicate which, if any, priority schemes are supported,
```
bslassey: the intent of this negotiation is also to allow the peers to specify their ranked preferences of prioritization schemes, so we should note that here.

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,62 @@ The terms sh-token and sh-boolean are imported from
...
+including the priority scheme as defined in [RFC7540], Section 5.3.
+
+For both HTTP/2 and HTTP/3, either peer's SETTINGS may arrive first,
+so any negotiation must be unilateral and not rely upon receiving
+the peer's SETTINGS value.
+
+Implementations are likely to only use one prioritization scheme at
+once, so the SETTINGS SHOULD be sent prior to the first request.
```
bslassey: for H2 it should probably be a MUST that the setting is sent before the first request

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,62 @@ The terms sh-token and sh-boolean are imported from
...
+
+The new parameter name is SETTINGS_PRIORITIES, which allows both
+peers to indicate which prioritization schemes they support.
+
+A value of 0 indicates no support for priorities. If either side sends the
+parameter with a value of 0, clients SHOULD NOT send priority frames
+and servers SHOULD NOT make any assumptions based on the presence or
+lack thereof of priority frames.
+
+If the value is non-zero, then the least significant 8 bits indicates the
```
bslassey, line 143: What if the 8 least significant bits are 0 but the next 8 bits are not?

ianswett: Added below.

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+all more significant octets MUST also be 0.
+
+In HTTP/2, the setting SHOULD appear in the first SETTINGS frame and peers
+MUST NOT process the setting if it's received multiple times in order to
+avoid changing the agreed upon prioritization scheme.
+
+If there is a prioritization scheme supported by both the client and server,
+then the servers's preference order prevails and both peers SHOULD
+only use the agreed upon priority scheme for the remainder of the session.
+The server chooses becasue it is in the best position to know what
```
rmarx:
```suggestion
The server chooses because it is in the best position to know what
```

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+information from the client is of the most value.
+
+An 8 bit value of 1 in HTTP/2 indicates support for HTTP/2 priorities
+as defined in Section 5.3 of [RFC7540] and is an error in HTTP/3 because
+there is not a clean mapping to HTTP/3.
+
```
rmarx, line 162: What value indicates the scheme defined in this document?

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+once, and may be unable to change the scheme once established, so the
+setting MUST be sent prior to the first request if it is ever sent.
+In HTTP/3, SETTINGS may arrive after the first request even if
+they are sent first.  In order to avoid the server incorrectly
+choosing a priority scheme in HTTP/3, the client SHOULD only send
+priority information prior to receiving the server's SETTINGS if
+it only offers a single scheme.  If it offers multiple schemes,
+the client SHOULD delay sending priority information until
+after it knows what the server supports, or risk the server
+choosing a sub-optimal scheme.
```
rmarx, line 128: Wondering if this SHOULD should be a MUST... if not, we might want to specify what a server is to do when it receives multiple, conflicting indicators (i.e., with this definition, it will start using the first one it sees, thinking it's the only one the client supports).

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+A value of 0 indicates no support for priorities. If either side sends the
+parameter with a value of 0, clients SHOULD NOT send priority frames
+and servers SHOULD NOT make any assumptions based on the presence or
+lack thereof of priority frames.
+
+If the value is non-zero, then the least significant 8 bits indicates the
+peer's preferred priority scheme, the second least significant 8 bits
+indicates the peer's second choice, and so on.  This allows expressing
+support for 4 schemes in HTTP/2 and 7 in HTTP/3.  If any octet is 0,
+all more significant octets MUST also be 0.
```
rmarx, line 147: It's difficult to look ahead and know how many schemes will exist, but I'm afraid 4 (or even 7) might be low in the long run. It depends on if "schemes" will be mutually exclusive or not. This is the case for the current + H2's dep tree, but if we go with the header-based design, I can see later schemes being extensions on the current one. See for example #28 or #50. 

Put differently: If we later specify new header values, like "blocking" or "visible", in a separate draft that extends upon this one, would that be a new value? And how does that work with negotiating if a peer only supports this current draft (which bisects nicely with the new one)

Or is the "scheme" more like the "delivery method", where it indicates header-based priorities instead of specific header values/options that may be contained therein...

LPardue: This design restricts the size of the presented list but not the size of the code space. 8-bits give enough(maybe?) codepoints while allowing a reasonable amount of room for advertising support for multiple priority schemes

draft-kazuho-httpbis-priority.md
LPardue: That's on one of @Kazuho's PRs butnwill need to be updated to reflect the depreciation of draft-lassey

LPardue:
When treating the settings parameter value as a list, we need also to describe how to handle duplicates. I suggest duplicates are an error but that we special case zero; something like treating 8-bit zero as a null character terminator, so presenting all zeroes is effectively an empty list.

draft-kazuho-httpbis-priority.md
rmarx: I agree with that. I just wonder: if current document is value 2, and follow-up is 3 (which includes 2 and adds new values): if client sends 3 but server only has 2, what happens? If client also has to send 2 besides 3, then I can see problems down the line with only allowing 4-7 discrete values, depending on how fine-grained new definitions will be in the future. 

Hence, we allow 8 bits so we can have -many- schemes in the future. But we can only select 4-7  of those at the same time, kind of undoing/restricting the usefulness of many schemes?

draft-kazuho-httpbis-priority.md
LPardue: I think the tradeoff is reasonable given the constraints. 

Trying to join up lists over multiple SETTINGS frames, or even multiple settings parameters is not something I would like to implement. 

Using a Frame allows as much space as needed, I spaced it out in an unshared doc, but I dont like the additional work it adds and the cost of losing support for remembered SETTINGSs

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+priority information prior to receiving the server's SETTINGS if
+it only offers a single scheme.  If it offers multiple schemes,
+the client SHOULD delay sending priority information until
+after it knows what the server supports, or risk the server
+choosing a sub-optimal scheme.
+
+## The SETTINGS_PRIORITIES SETTINGS Parameter
+
+This document adds a new SETTINGS parameter to those defined by
+[RFC7540], Section 6.5.2.
```
LPardue, line 133: I'd combine the para with the next and reword to say

 "This document defines a new SETTINGS_PRIORITIES parameter for HTTP/2 and HTTP/3, which allows both peers to indicate which prioritization schemes they support."

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+MUST NOT process the setting if it's received multiple times in order to
+avoid changing the agreed upon prioritization scheme.
+
+If there is a prioritization scheme supported by both the client and server,
+then the servers's preference order prevails and both peers SHOULD
+only use the agreed upon priority scheme for the remainder of the session.
+The server chooses because it is in the best position to know what
+information from the client is of the most value.
+
+An 8 bit value of 1 in HTTP/2 indicates support for HTTP/2 priorities
```
LPardue, line 159:
```suggestion
An 8-bit value of 1 in HTTP/2 indicates support for HTTP/2 priorities
```

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+as defined in Section 5.3 of [RFC7540] and is an error in HTTP/3 because
+there is not a clean mapping to HTTP/3.
```
LPardue, line 161: I think I'd like to see values defined a little more formally. For example, a proper noun for these things (priorty scheme identifier?), a human-friendly identifer name (that we can bikeshed on) and a codepoint. Then I would restate things as:

"This document defines the priority scheme identifier H2_TREE (8-bit value of 1) that indicates support for HTTP/2-style priorities. 

The H2_TREE priority scheme identifier MUST NOT be be sent in an HTTP/3 settings because there is no defined mapping of this scheme. Endpoints MUST treat receipt of H2_TREE as a connection error of type H3_SETTINGS_ERROR."

I also think there is value in an IANA registry for these identifiers.

draft-kazuho-httpbis-priority.md
LPardue: Also need to pick a code point and ideally update this document's IANA section.

draft-kazuho-httpbis-priority.md
kazuho: IMO, there are two problems with the text.

One is that it might be too restrictive. We do not know what other priority schemes we'd see in the future. Consider the case where we'd have a frame-based prioritization scheme that is dedicated to video requests, while the other streams would be using the Priority headers. In such case, there is no reason to prohibit either of the two from being sent until the client learns that the SETTINGS have been received.

The other problem is that H3 does not define a way to acknowledge the receipt of SETTINGS at the application-protocol level. It is my understanding that H3 (including QPACK) is designed to have it's own ACKs for signaling the receipt, rather than relying on transport-level ACKs. This normative text violates that principle.

Considering these aspects, I think that we should paraphrase this these sentences discussing HTTP/3 to something like: _In HTTP/3, SETTINGS may arrive after the first request even if they are sent first. Future specifications that define alternative prioritization schemes SHOULD define how the server would act when it receives stream-level priority signals prior to receiving the SETTINGS frame_.

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+The new parameter name is SETTINGS_PRIORITIES, which allows both
+peers to indicate which prioritization schemes they support.
+
+A value of 0 indicates no support for priorities. If either side sends the
+parameter with a value of 0, clients SHOULD NOT send priority frames
+and servers SHOULD NOT make any assumptions based on the presence or
+lack thereof of priority frames.
+
+If the value is non-zero, then the least significant 8 bits indicates the
+peer's preferred priority scheme, the second least significant 8 bits
```
kazuho, line 144: Isn't this the preferred priority scheme of the sender (the endpoint) than the peer? I think we might also want to say "MOST".

In total, something like "the sender's most preferred priority scheme."

draft-kazuho-httpbis-priority.md
```diff
@@ -98,6 +98,70 @@ The terms sh-token and sh-boolean are imported from
...
+support for 4 schemes in HTTP/2 and 7 in HTTP/3.  If any octet is 0,
+all more significant octets MUST also be 0.
+
+In HTTP/2, the setting SHOULD appear in the first SETTINGS frame and peers
+MUST NOT process the setting if it's received multiple times in order to
+avoid changing the agreed upon prioritization scheme.
+
+If there is a prioritization scheme supported by both the client and server,
+then the servers's preference order prevails and both peers SHOULD
+only use the agreed upon priority scheme for the remainder of the session.
```
kazuho, line 155: Would it make sense to state
* that "the server SHOULD only use the agreed upon priority scheme",
* and that "the endpoints SHOULD refrain from sending frame-level signals that were not agreed"?

The rationale behind the latter is in the last paragraph of https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/55/files.

kazuho (commented):
Thank you for the PR. Looks mostly good, please see my comments below.

LPardue:
This PR is mostly there, and is now blocking some other things, so in the interest of making progress before the deadline I'm going to merge/tidy up

kazuho:
Merged as part of #61.

ianswett:
@LPardue Thanks for fixing it up and merging!

Issue #32, opened 2019-07-10
Describe precedence more


LPardue:
We use the term "absolute precedence" or "precedence" in several places in the I-D. So it seems special but IMO It doesn't leap of the page what this means, although we hint towards it in places such as:

> A server SHOULD transmit HTTP responses in the order of their urgency: blocking first, followed by document, followed by non-blocking.

and 

> A server SHOULD transmit non-progressive responses one by one, preferably in the order the requests were generated.

LPardue:
@kazuho while revisiting this issues I noticed the present draft could be a little confusing. In the abstract we say 

> The prioritization scheme expresses absolute values for each HTTP response, rather than the relative precedence

This might imply some difference between values and precedence - but then every instance of precedence in the main document seems to refer to usage as a value

kazuho:
@LPardue Please see #60, or concrete suggestions are very welcome.

LPardue:
Thanks for the abstract fix. Let me continue to think on the precedence thing a little more in the main document. If I cant suggest something by end of Singaport I'll close the issue.

PR #60, opened 2019-11-03
Do not use "precedence" (at least in abstract)

kazuho:
@LPardue Is this better (and flow well)?

See #32 for discussion.


LPardue (approved):
Awesome

PR #61, opened 2019-11-03
Consolidation of the priority problem statement

LPardue:
On the priority team mailing list I advocated for a [consolidation of the problem statement](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/http-priorities-design-team/an06MqgKgIE)

This PR builds on top of #58, creating a new section that combines that text, with text from the introduction and pulls in some of Roberto's text. All of that happens on https://github.com/kazuho/draft-kazuho-httpbis-priority/commit/b2b927ccbeea76f9edcb5e49ad9be06530d9e2ab

@rmarx I didn't pull anything specifically from your paper but if you feel there is something that I don't cover then please do suggest it

PR #73, opened 2019-11-04
Expand motivation to include the choice for a simpler scheme

rmarx:
As mentioned in https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/61#issuecomment-549329078

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -126,9 +126,10 @@ resource exhaustion vectors affecting multiple HTTP/2 implementations. One
 attack, CVE-2019-9513 aka "Resource Loop", is based on manipulation of the
 priority tree.
 
-The HTTP/2 scheme depends on in-order delivery, so it is unsuitable for direct
-use in protocols like HTTP/3 {{?I-D.ietf-quic-http}}, which attempts to avoid
-global ordering.
+The HTTP/2 scheme depends on in-order delivery of signals, leading to challenges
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -125,18 +126,21 @@ resource exhaustion vectors affecting multiple HTTP/2 implementations. One
 attack, CVE-2019-9513 aka "Resource Loop", is based on manipulation of the
 priority tree.
 
-The HTTP/2 scheme depends on in-order delivery, so it is unsuitable for use in
-protocols like HTTP/3 {{?I-D.ietf-quic-http}}, which attempts to avoid global
-ordering.
+The HTTP/2 scheme depends on in-order delivery, so it is unsuitable for direct
```
LPardue: I don't think `direct` is the best term to use here, I understand the meaning because I have the context.

How about we rework the paragraph a bit as so

"The HTTP/2 scheme depends on in-order delivery of signals, there are challenges in porting the scheme to protocols that do not provide global ordering. For example, the scheme cannot be used in HTTP/3 {{?I-D.ietf-quic-http}} without changing the signal and its processing."

LPardue (approved):
looks good, modulo the comment I made

draft-kazuho-httpbis-priority.md
rmarx: Changed in 6e95bd0

kazuho:
Merged! Thank you for the PR and the review.

PR #74, opened 2019-11-04
Update reprioritization example

### LINKED ISSUE #53: Reprioritization example of the HTML file a bit unrealistic maybe
I feel the reprioritization example of the HTML file a bit unrealistic maybe. Are people actually prefetch-ing whole documents? I personally would change it to a prefetch-ed JavaScript request or maybe an image that was off-screen but now the user started scrolling down.  

_Originally posted by @rmarx in https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/51#issuecomment-548138752_

rmarx:
Fixes #53

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -446,7 +446,7 @@ might issue a prefetch request for a JavaScript file with the urgency parameter
 of the Priority request header field set to `urgency=6` (background). Then, when
 the user navigates to a page which references the new JavaScript file, while the
 prefetch is in progress, the browser would send a reprioritization frame with the
-priority field value set to `urgency=0` (default).
+priority field value set to `urgency=-1` (prerequisite).
 
 However, a client cannot reprioritize a response by using the Priority header
 field.  This is because an HTTP header field can only be sent as part of an HTTP
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -442,10 +442,11 @@ server and the intermediary.
 
 Once a client sends a request, circumstances might change and mean that it is
 beneficial to change the priority of the response. As an example, a web browser
-might issue a prefetch request for an HTML document with the urgency parameter
+might issue a prefetch request for a JavaScript file with the urgency parameter
 of the Priority request header field set to `background`. Then, when the user
-navigates to the HTML while prefetch is in action, it would send a
-reprioritization frame with the priority field value set to `urgency=0`.
+navigates to a page which references the new JavaScript file, while the prefetch
+is in progress, the browser would send a reprioritization frame with the priority
+field value set to `urgency=0`.
 
 However, a client cannot reprioritize a response by using the Priority header
 field.  This is because an HTTP header field can only be sent as part of an HTTP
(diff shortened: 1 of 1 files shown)


LPardue (changes requested):
This looks good, but I made a stylistic suggestion that I think will help readability.

rmarx:
I'm not sure what you mean... I can't seem to see the "stylistic suggestion" using the github interface for some reason. However, I'm sure it's fine and you can just merge afaic.

draft-kazuho-httpbis-priority.md
```diff
@@ -442,10 +442,11 @@ server and the intermediary.
 
 Once a client sends a request, circumstances might change and mean that it is
 beneficial to change the priority of the response. As an example, a web browser
-might issue a prefetch request for an HTML document with the urgency parameter
+might issue a prefetch request for a JavaScript file with the urgency parameter
 of the Priority request header field set to `background`. Then, when the user
-navigates to the HTML while prefetch is in action, it would send a
-reprioritization frame with the priority field value set to `urgency=0`.
+navigates to a page which references the new JavaScript file, while the prefetch
```
kazuho: Assuming that the example talks about SPA, is there a "page" navigation?

I do not mind merging the PR as-is, and I do not know about SPA, but I am just wondering if there is a technical error in the example.

LPardue:
@rmarx Lol Github ate my suggestion. To sumamrise, the paragraph mixes priority level names with values e.g. it uses `background` and then a reprioritiziation to `urgency=0`. I think it would help to use both in this example. I.e. after `background` add "(`urgency=6`)", and then similarly have "`default` (`urgency=1`)

draft-kazuho-httpbis-priority.md
rmarx: I agree, but a) this works for a normal page navigation as well (you can prefetch files for the next full page load) and b) in the context of an SPA, we typically still speak of a "navigation" even though it doesn't include a full page load (e.g., the URI is updated, browser history is appended, etc.). 

I felt it was easier to just keep "navigate" than to start talking about the SPA case in particular.

kazuho: Thank you for the explanation. I think I'm fine with the text.

rmarx:
Aha, excellent suggestion!
Changed in f6604a1

draft-kazuho-httpbis-priority.md
```diff
@@ -442,10 +442,11 @@ server and the intermediary.
...
 beneficial to change the priority of the response. As an example, a web browser
-might issue a prefetch request for an HTML document with the urgency parameter
-of the Priority request header field set to `background`. Then, when the user
-navigates to the HTML while prefetch is in action, it would send a
-reprioritization frame with the priority field value set to `urgency=0`.
+might issue a prefetch request for a JavaScript file with the urgency parameter
+of the Priority request header field set to `urgency=6` (background). Then, when
+the user navigates to a page which references the new JavaScript file, while the
+prefetch is in progress, the browser would send a reprioritization frame with the
+priority field value set to `urgency=0` (default).
```
kazuho: Should the urgency be -1, assuming that this JavaScript file is render-blocking?

rmarx: I think both are fine, since JS is not necessarily render blocking, not even if it's prefetched. Yet, usually it is, and I feel it's best to encourage people to give JS high priorities, so updated in fad2b3c.

LPardue (approved):

ianswett (approved):
LG, though it reminds me we never changed the range to be 0 to 7.

kazuho:
Thank you all!!!

PR #76, opened 2019-11-04
"priority frames" is ambiguous, talk about hop-by-hop signals

kazuho:
...that could be a H2 PRIORITY frame or a payload of a HEADERS frame or whatever.


LPardue (approved):
Good spot and fix

PR #79, opened 2019-11-04
hop-by-hop signals may cease once negotiation is complete but end-to-end should not

### LINKED ISSUE #62: Scheme negotiation normative language
Would it make sense to state
* that "the server SHOULD only use the agreed upon priority scheme",
* and that "the endpoints SHOULD refrain from sending frame-level signals that were not agreed"?

The rationale behind the latter is in the last paragraph of https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/55/files.

_Originally posted by @kazuho in https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/58_

kazuho:
Suggests that hop-by-hop signals might cease, clarifies more generally that end-to-end signals should not.

Closes #62.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -195,6 +195,12 @@ only use the agreed upon priority scheme for the remainder of the session.
 The server chooses because it is in the best position to know what
 information from the client is of the most value.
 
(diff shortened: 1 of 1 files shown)


LPardue (approved):
lgtm

PR #80, opened 2019-11-04
Split prioritization into scheme and header sections

ianswett:
I added a bit of text to introduce the 'Urgency' prioritization scheme, but besides that this is just moving text around.  I think the header section now needs more explanation, but that can be added in a different PR.

I'm happy to pick a different name from 'Urgency' as well, but it seemed like the most obvious choice.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -195,10 +195,10 @@ only use the agreed upon priority scheme for the remainder of the session.
 The server chooses because it is in the best position to know what
 information from the client is of the most value.
 
-Once the negotiation is complete, endpoints MAY stop sending hop-by-hop	
-prioritization signals that were not negotiated in order to conserve bandwidth.	
-However, endpoints SHOULD continue sending end-to-end signals (e.g., the	
-Priority header field), as that might have meaningful effect to other nodes that	
+Once the negotiation is complete, endpoints MAY stop sending hop-by-hop
+prioritization signals that were not negotiated in order to conserve bandwidth.
+However, endpoints SHOULD continue sending end-to-end signals (e.g., the
+Priority header field), as that might have meaningful effect to other nodes that
 handle the HTTP message.
 
 An 8 bit value of 1 in HTTP/2 indicates support for HTTP/2 priorities
@@ -217,7 +217,7 @@ has negotiated the use of the extensible priority scheme (see {{fairness}}).
 
 # Urgency Prioritization
 
-Urgency based prioritization uses two fields, `urgency` and `progressive` to
+Urgency prioritization uses two parameters, `urgency` and `progressive` to
 indicate the relative urgency of requests and whether responses should be
 interleaved with other responses at the same priority.
 
@@ -340,8 +340,6 @@ uses it to specify the priority of the response. A server uses it to inform
 the client that the priority was overwritten. An intermediary can use the
 Priority information from client requests and server responses to correct or
 amend the precedence to suit it (see {{merging}}).
-
-Unknown parameters MUST be ignored.
 
 
 ## urgency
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -210,20 +204,23 @@ there is not a clean mapping to HTTP/3.
 The extensible priority scheme is negotiated using the described mechanism. It is
 identified by the 8-bit value of 2.
 
+Endpoints using this scheme SHOULD emit the Priority header field regardless of
```
ianswett: I think this PR accidentally reverted this change?  I can fix that....

kazuho: I think that is the case.

draft-kazuho-httpbis-priority.md
```diff
@@ -210,20 +204,23 @@ there is not a clean mapping to HTTP/3.
...
 
-# The Priority HTTP Header Field
+# Urgency Prioritization
 
-The Priority HTTP header field can appear in requests and responses. A client
-uses it to specify the priority of the response. A server uses it to inform
-the client that the priority was overwritten. An intermediary can use the
-Priority information from client requests and server responses to correct or
-amend the precedence to suit it (see {{merging}}).
+Urgency based prioritization uses two fields, `urgency` and `progressive` to
```
kazuho: I think you might want to change "fields" to "parameters", as that is the term we use throughout.

And it might be a good idea to change the section title to indicate that the section discusses various parameters collectively, though I do not have a good suggestion.

draft-kazuho-httpbis-priority.md
```diff
@@ -345,6 +331,33 @@ order the requests were generated.  Doing so maximizes the chance of the client
...
+
+# The Priority HTTP Header Field
+
+The Priority HTTP header field can appear in requests and responses. A client
+uses it to specify the priority of the response. A server uses it to inform
+the client that the priority was overwritten. An intermediary can use the
+Priority information from client requests and server responses to correct or
+amend the precedence to suit it (see {{merging}}).
+
+Unknown parameters MUST be ignored.
```
kazuho: I think this belongs to the section above.

kazuho (commented):
Thank you for the PR.

While I can see the intent of the reorganization, I feel sad that we are losing examples. I tend to think that talking about the version-independent design, then going into how to do version-specific extensions (i.e. reprioritization) might help the readers.

draft-kazuho-httpbis-priority.md
ianswett: Thanks, I changed fields to parameters.

The intent was to discuss this as a priority scheme, not just a list of parameters, but maybe there's a better way to do it.

draft-kazuho-httpbis-priority.md
ianswett: Thanks, removed.

LPardue:
@ianswett I think there is more work required to make this all hold together. I  made a more extensive comment on possible improvements and GitHub lost it. How critical do you feel it is for this to be included in the draft?

ianswett:
I'd like to split the prioritization scheme from how its conveyed, since I think it's good to separate concerns, but it doesn't have to be done in the way I did it in this PR.

I'd also like more detail on how the header works(ie: is it cacheable), in this section, opposed to some text in "Why use an End-to-End Header Field?" but we're running low on time for that.

draft-kazuho-httpbis-priority.md
ianswett: Reverted.

ianswett:
After re-reading your comments, I think we're thinking about how this works somewhat differently.

I'm thinking of the core of the proposal being the scheme, and then there are two ways(header and frame) to convey that information.  Maybe you're thinking the core is the header itself?

LPardue:
I think your concept could work but the current text doesn't do that well. We would need to talk about scheme parameters of urgency (an integer,) and progressive (a bool) in abstract sense.  Parameters are extensible and unknown one must be ignored.

Then we can talk about specific encoding of parameters, which is structured headers dictionary members: sh-integer and sh-boolean. And that the rlencoded dictionary of parameters is what gets sent.

The text is not quite there in splitting these out

LPardue:
Closing "split"-related candidate changes because they have rotted. We may chose to address the goal of these changes, and we may chose to restructure the document, and we may chose to do both of those things together; but I don't think this PR helps.

PR #81, opened 2019-11-04
Add formal scheme identifier names and IANA table

LPardue:
Fixes #56 by being more explicit about the 8-bit values that are used inside the new SETTINGS_PRIORITIES parameters.

It goes "whole hog" and claims a SETTINGS value rather than leaving it TBD (but i'm happy to roll that back if people feel strongly against it)

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -162,21 +162,21 @@ stream-level priority signal prior to receiving the SETTINGS frame.
 
 ## The SETTINGS_PRIORITIES SETTINGS Parameter
 
-This document adds a new SETTINGS parameter to those defined by
-[RFC7540], Section 6.5.2.
+This document defines a new SETTINGS_PRIORITIES parameter (0x9) for HTTP/2 and
+HTTP/3, which allows both peers to indicate which prioritization schemes they
+support. The value of this parameter is interpreted in two ways depending on if it is
+zero or non-zero.
 
-The new parameter name is SETTINGS_PRIORITIES, which allows both
-peers to indicate which prioritization schemes they support.
+If the setting has a value of zero it indicates no support for priorities. If
+either side sends the parameter with a value of zero, clients SHOULD NOT send
+hop-by-hop priority signals (e.g., HTTP/2 PRIORITY frame) and servers SHOULD NOT
+make any assumptions based on the presence or lack thereof of such signals.
 
-A value of 0 indicates no support for priorities. If either side sends the
-parameter with a value of 0, clients SHOULD NOT send hop-by-hop prioritiy
-signals (e.g., HTTP/2 PRIORITY frame) and servers SHOULD NOT make any
-assumptions based on the presence or lack thereof of such signals.
-
-If the value is non-zero, then the least significant 8 bits indicate the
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -201,14 +201,25 @@ However, endpoints SHOULD continue sending end-to-end signals (e.g., the
...
+that indicates support for HTTP/2-style priorities ({{!RFC7540}}, Section 5.3).
+
+The H2_TREE priority scheme identifier MUST NOT be be sent in an HTTP/3 settings
+because there is no defined mapping of this scheme. Endpoints MUST treat receipt
+of H2_TREE as a connection error of type H3_SETTINGS_ERROR.
+
+### EXT_PRI {#settings-this-scheme}
+
+This document defines the priority scheme identifier EXT_PRI (8-bit value of 2)
+that indicates support for the extensible priority scheme defined by the in the
```
kazuho: "by the" or "in the"?

kazuho (approved):
Thank you for the PR. LGTM modulo the point below.

LPardue:
CC @ianswett too for some review

draft-kazuho-httpbis-priority.md
LPardue: thanks, fixed in https://github.com/kazuho/draft-kazuho-httpbis-priority/pull/81/commits/94828cd02651252c78abf9b5f663048a0486adb7

draft-kazuho-httpbis-priority.md
```diff
@@ -201,14 +201,25 @@ However, endpoints SHOULD continue sending end-to-end signals (e.g., the
...
-The extensible priority scheme is negotiated using the described mechanism. It is
-identified by the 8-bit value of 2.
+This document defines the priority scheme identifier H2_TREE (8-bit value of 1)
+that indicates support for HTTP/2-style priorities ({{!RFC7540}}, Section 5.3).
+
+The H2_TREE priority scheme identifier MUST NOT be be sent in an HTTP/3 settings
+because there is no defined mapping of this scheme. Endpoints MUST treat receipt
+of H2_TREE as a connection error of type H3_SETTINGS_ERROR.
+
+### EXT_PRI {#settings-this-scheme}
```
ianswett: I'd call this EXTENSIBLE.  Including _PRI is redundant to the SETTING name.

I also don't think this is the name I'd choose, because if you extended the current scheme to add a new field, I believe that needs its own value?

ianswett (approved):
Thanks for adding this.  I don't think calling the scheme extensible tells you much about what it offers in terms of functionality, but happy to discuss that on a separate issue.

draft-kazuho-httpbis-priority.md
kazuho: Then maybe ABSOLUTE?

Regardless, I think we can leave it as-is (or whatever) because it's does not affect the logic.

LPardue: Good points. If we applied @ianswett change on #80 the scheme would be called "Urgency Prioritization", so how about `URGENCY` as a replacement for `EXT_PRI`?

kazuho: Works for me.

PR #85, opened 2019-11-04
Split tweaking

LPardue:


LPardue:
This builds in top of #80 to implement the tweaks I had in mind. @ianswett PTAL. Since it is getting close to deadline, if this is too contentious we might need to hold it back until next time.

draft-kazuho-httpbis-priority.md
```diff
@@ -250,17 +252,6 @@ The value is encoded as an sh-integer.  The default value is zero.
...
-`-1`:
-
-~~~ example
-:method = GET
-:scheme = https
-:authority = example.net
-:path = /style.css
-priority = urgency=-1
-~~~
-
```
kazuho: I think having examples here would greatly help readers internalize how it works as they read the text. We can have a back reference to the header definition.

draft-kazuho-httpbis-priority.md
```diff
@@ -345,6 +336,31 @@ order the requests were generated.  Doing so maximizes the chance of the client
 making progress in using the composition of the HTTP responses at the earliest
 moment.
```
kazuho: We should move the example of the progressive parameter to here, too.

draft-kazuho-httpbis-priority.md
```diff
@@ -345,6 +336,31 @@ order the requests were generated.  Doing so maximizes the chance of the client
...
+# The Priority HTTP Header Field {#header-field}
+
+The Priority HTTP header field can appear in requests and responses. A client
+uses it to specify the priority of the response. A server uses it to inform
+the client that the priority was overwritten. An intermediary can use the
+Priority information from client requests and server responses to correct or
+amend the precedence to suit it (see {{merging}}).
+
+
+## urgency
```
kazuho: I do not think we need a sub-section just for each example.

Instead, assuming that we would have examples above, just having one another example in this section as a whole would be fine.

kazuho (commented):
Thank you for the PR. The text looks fine, though the examples seem not to be in the best places.

LPardue:
So after juggling the examples (because I agree with Kazuho about concrete examples) the "split" isn't really a split. I think we probably want to follow up in a 3-way discussion to see what exactly we are trying to achieve and see if it makes sense to land a change in an 04 draft at a later date.

draft-kazuho-httpbis-priority.md
```diff
@@ -215,23 +215,25 @@ extensible priority scheme when and only when all the requests that are to be
 sent on that backend connection originates from one client-side connection that
 has negotiated the use of the extensible priority scheme (see {{fairness}}).
 
-# The Priority HTTP Header Field
+# Urgency Prioritization
```
kazuho, line 218: I'm starting to think that this name is confusing.

In my view, this section discusses two things: 1) the extensible scheme using the Structured Headers, 2) the urgency and the progressive parameters.

Even if we are to consider calling (2) as "Urgency Prioritization" to be okay, I think it's wired to refer to 1 using the same name.

ianswett: Agreed, including the structured header part in this section is a bit odd.

ianswett:
A 3 way discussion sounds good to me.  I'll note that we have extensibility in the form of negotiation via SETTINGS, so I think I'd like a clearer picture of what types of changes should use SETTINGS and what should use another field in the structured header.

My thinking was that we should have a SETTINGS codepoint for the design in this draft(urgency and progressive) and if one wanted to have a scheme that used urgency and some to be defined field, that would be a separate codepoint.

LPardue:
So a way to use structured headers encoded parameters but without the mandate to use urgency and progressive parameters (or imply behaviour from their absence)

draft-kazuho-httpbis-priority.md
kazuho: Reconsidering this, I think we need to agree on what we expose as layers, based on how we think that the extensibility is going to be used. I've opened #86. PTAL.

LPardue:
Closing "split"-related candidate changes because they have rotted. We may chose to address the goal of these changes, and we may chose to restructure the document, and we may chose to do both of those things together; but I don't think this PR helps.

PR #103, opened 2019-11-19
The header is end-to-end

### LINKED ISSUE #45: Should the priority header be hop-by-hop?
Making the priority header end-to-end seems like a large departure from HTTP/2 priorities and one I'm not sure we have much/any implementation experience with.  In particular, I know many load balancers and CDNs are still using HTTP 1.1 to backends/application front ends.  In the case of HTTP 1.1, it seems like the intent is to inform the AFE of the request priority and allow overriding it, rather than having the priority apply to the request itself?

It's fairly easy to understand what an implementation should do with a hop-by-hop priority header of this sort, but if a request goes through a load balancer and is proxied to a backend and pooled with other requests, it's not clear to me that the request should be treated as higher priority than the other requests?

One example that sticks out to me is if there was a 'tt'l key in the request header indicating to the server: "Don't bother sending the response after this time", then carrying that to another server would be potentially misleading, since an unknown time would have elapsed between the request being sent and it reaching the second server.  In that case, it seems like the load balancer should make a best effort at adjusting the ttl.

If they are going to remain end-to-end, I think more explanation of what that means is important and a few examples would be helpful.  Currently, most or all of the relevant text is in the "Considerations" section.

### LINKED ISSUE #57: Hop-by-hop emulation with end-to-end headers
Not necessarily an issue; more that I wanted to sketch out the idea I had during the call a few weeks ago.

The application of any prioritization is within the context of either the connection or the server itself.  The server's cross-connection prioritization is open enough to gaming (my requests are always more important than yours!) that I feel like we should specifically advise against it and design solely for intra-connection prioritization.

I think there are three key values to be transported across each hop:

- HTTP-Priority (C2S):  Expresses the client's preference for how the response is handled on this connection.
- Original-HTTP-Priority (C2S):  Carries the user agent's preference for how the response is handled on the user-to-first-intermediary connection (which is not this one).
- HTTP-Priority (S2C):  Expresses the server's preference for how the response is handled on downstream connections, if any

Processing rules would be that each of these headers is end-to-end, with one key exception: intermediaries MAY replace the HTTP-Priority value with their own preferences when making upstream requests.  If they do so and there isn't an Original-HTTP-Priority header already, they SHOULD copy the original value to the Original-HTTP-Priority header.  An intermediary that wants to use this scheme upstream, but did not receive an HTTP-Priority header from the client should add an "Original-HTTP-Priority: (none)" header field.

There's a fourth value that is probably not worth putting on the wire, which is what the server (on a given connection) *has actually chosen to do* on this connection.  That's an implementation artifact, and it doesn't need to get shared outside its code.  The S2C header is useful, however, if the server wants to override what the client said *or* provide cachable guidance about how a cached version of this response should be sent to clients in the future, even in the absence of guidance from that future client.

Each intermediary is looking at the suggested policies from its client (which might not be the user agent) and from the server and choosing what policy to apply in an implementation-dependent manner.

ianswett:
The frame is hop-by-hop

Fixes #45 and fixes #57

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -242,7 +242,7 @@ defined parameters are omitted, or if the Priority header field is not used,
 their default values SHOULD be applied.
 
 The Priority header field is an end-to-end signal of the request
-client or the response priority from the server.
+priority from the client or the response priority from the server.
 
 Unknown parameters MUST be ignored.
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -241,7 +241,7 @@ parameters. Values of these parameters MUST always be present. When any of the
 defined parameters are omitted, or if the Priority header field is not used,
 their default values SHOULD be applied.
 
-The priority header is an end-to-end signal of the request priority from the
+The Priority header field is an end-to-end signal of the request
 client or the response priority from the server.
 
 Unknown parameters MUST be ignored.
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -241,6 +241,9 @@ parameters. Values of these parameters MUST always be present. When any of the
 defined parameters are omitted, or if the Priority header field is not used,
 their default values SHOULD be applied.
 
+The priority header is an end-to-end signal of the request priority from the
+client or the response priority from the server.
+
 Unknown parameters MUST be ignored.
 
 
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -241,6 +241,9 @@ parameters. Values of these parameters MUST always be present. When any of the
 defined parameters are omitted, or if the Priority header field is not used,
 their default values SHOULD be applied.
 
+The priority header is an end-to-end signal of the request priority from the
```
```suggestion
The Priority header field is an end-to-end signal of the request
```

use `field` and reflow text

draft-kazuho-httpbis-priority.md
```diff
@@ -241,6 +241,9 @@ parameters. Values of these parameters MUST always be present. When any of the
...
+client or the response priority from the server.
```
```suggestion
priority from the client or the response priority from the server.
```

draft-kazuho-httpbis-priority.md
```diff
@@ -397,6 +400,7 @@ avoided by restricting the stream on which a PRIORITY_UPDATE frame can be sent.
 In HTTP/2 the frame is on stream zero and in HTTP/3 it is sent on the control
 stream ({{!I-D.ietf-quic-http}}, Section 6.2.1).
 
+Unlike the header, the reprioritization frame is a hop-by-hop signal.
```
```suggestion
Unlike the header field, the reprioritization frame is a hop-by-hop signal.
```

LPardue (approved):

PR #104, opened 2019-11-19
Add a section on Extensible Priorities

ianswett:
With an initial set of text.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -248,12 +248,10 @@ the client that the priority was overwritten. An intermediary can use the
 Priority information from client requests and server responses to correct or
 amend the precedence to suit it (see {{merging}}).
 
-The value of the Priority header field is a Structured Headers Dictionary
-({{!STRUCTURED-HEADERS}}). Each dictionary member represents a parameter of the
-Priority header field. This document defines the `urgency` and `progressive`
-parameters. Values of these parameters MUST always be present. When any of the
-defined parameters are omitted, or if the Priority header field is not used,
-their default values SHOULD be applied.
+This document defines the `urgency` and `progressive` parameters. Values of
+these parameters MUST always be present. When any of the defined parameters
+are omitted, or if the Priority header field is not used, their default
+values SHOULD be applied.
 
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -226,6 +226,20 @@ extensible priority scheme when and only when all the requests that are to be
...
 has negotiated the use of the extensible priority scheme (see {{fairness}}).
 
+# Extensible Priorities
+
+This priority design is intended to be extensible, and uses a sequence of
+key-value pairs encoded as a Structured Headers Dictionary
+({{!STRUCTURED-HEADERS}}) to enable extensibility. Each dictionary member
+represents a parameter of the Priority header field.
+
+Care must be taken to ensure any use of extisting parameters are either
```
```suggestion
Care must be taken to ensure any use of existing parameters are either
```

LPardue: I think the paragraph would also benefit from starting with the subject e.g. "When attempting to extend priorities, care must be taken.."

LPardue (changes requested):

LPardue (approved):
This is a good starting point, I suggest we merge it as-is in order to permit further progress.

kazuho:
Thank you for the text. I think that the clarification on extensibility is fine.

OTOH, this looks to me like an in-progress work of refactoring being merged.

The previous structure of the document was to first describe the header and the parameters, then state that the frame reuses that encoding.

This PR changed the structure to first describe that SH is used by headers and frame, then talk about the header field along with and parameters, then talk separately about frames.

I think we should either restore the original structure (requires minimal work), or go ahead and refactor the text so that we discuss parameters in a conveyer-independent way, then talk about headers and frames.

LPardue:
I'll take the blame here. I think a few people would appreciate a document restructure in order to make changes that are more content full and accurate

ianswett:
I was going to do the latter ("go ahead and refactor the text so that we discuss parameters in a conveyer-independent way, then talk about headers and frames.") this afternoon after outstanding PRs are merged.

kazuho:
I share the view that restructuring is preferable.

Issue #87, opened 2019-11-05
Explain why a client would send the priority header in HTTP 1.1


ianswett:
I can imagine a proxy using this to an origin in the CDN case, but especially if we're not going to suggest changing congestion control/etc (#48), I think the value of this header on an HTTP 1.1 connection is really low.

I don't believe we want priorities to influence cross-connection processing order?

LPardue:
I think this is a good issue to track.

I'll raise the Fetch card, web developers that programmatically generate requests have no control over the HTTP Protocol version that is used. So in a roundabout way for this case, the reason to send the priority header in HTTP/1.1 is because it meets developer expectations. 

For requests generated by a UA, such as initiated by a navigation, there may be benefit to omitting priorities in HTTP/1.1.

LPardue:
Misclicked close issue, sorry. I think we should keep this open until an answer is agreed upon.

c-taylor:
Does the expansion in #110 resolve this, If you ignore the editorial placement issue?

ianswett:
I'd suggest adding an editorial note that clients may want to omit the header on HTTP 1.1 as an optimization, but I think the text needs a restructure before I know where to put that exactly.

Issue #46, opened 2019-09-30
Priorities should affect request behavior as well as response behavior


ianswett:
Currently, there's no discussion of whether the priority header affects request behavior as well as response behavior.  I believe it should affect both whenever possible and that should be specified.

In some cases, that could be a browser delaying sending lower priority requests behind higher priority ones, and in others(ie: a large POST) that could be changing the congestion control characteristics to ensure the request would not significantly impact other requests.

kazuho:
:+1:

rmarx:
+1

LPardue:
The title "The Priority HTTP header field" makes these discussions hard because I'm not sure sure if @ianswett is talking about abstract absolute priority levels or concrete signal in the form of HTTP headers or frames. I think we should try to tackle terms in the next document update.

I don't understand how a priority header can affect the scheduling of requests inside the user agent. However, the abstract concept herein is that the UA generates the priority signal based on its local usage of the resource. So it is fully in the means of UAs to modulate their behaviour (ordering, or send rate) based on usage. Is that what is being proposed on this ticket? 

The actual recommended behaviour is likely to depend on the connection prioritization scheme. For example, signalling lower priorities later might work well for FIFO but would perform poorly for LIFO. How specific is the guidance that we are attempting to capture?

rmarx:
In my view, this is mostly for when developers are setting the priority themselves (e.g., indirectly using an HTML tag attribute or directly via fetch()). As we mentioned in the previous meeting, if "headers are the API" (much like how the preload header triggers push), then priority headers set by the developer on the client side (e.g., via JS) can/should trigger browser behavior changes.  

As Ian indicates, a developer performing a large background POST (e.g., from a service worker) could indicate this using priority headers, triggering appropriate browser-side behavior. I would see this as an option to impact HTTP/1.1 as well, where some requests can be delayed (not taking up 1 of the 6 allowed connections) if they are low priority.  

Granted, the benefits are not huge in my opinion, but enough to specifically adopt this use case (it's also nice and symmetrical ;))

LPardue:
As mentioned elsewhere, I'd like us to document use cases and actors so we can be clear in our communications.

[...]

PR #105, opened 2019-11-19
Describe the impact of priority on other connections

ianswett:
Priority is mostly within a connection, except when the server knows otherwise.

Fixes part of #87 and part of #46

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -240,6 +240,18 @@ compatible for peers that are unaware of the extended meaning.
 The scheme has a single encoding and set of functionality whether it's
 conveyed via a HTTP header or within a frame.
 
+## Impact on other connections
+
+When multiple connections are open, the priority MAY have impact on the
+client's local scheduling choices, but MUST NOT change how the server allocates
+resources among different incoming connections unless it has a guarantee some
+set of connections are from the same client. When a server knows multiple
+incoming connections are from the same client, the server MAY use the priority
+to inform scheduling decisions among the client's different connections.
+
+Due to this, priority information conveyed over a non-coalesced HTTP
+connection might go unused.
+
 # The Priority HTTP Header Field
 
 The Priority HTTP header field can appear in requests and responses. A client


draft-kazuho-httpbis-priority.md
```diff
@@ -240,6 +240,18 @@ compatible for peers that are unaware of the extended meaning.
 The scheme has a single encoding and set of functionality whether it's
 conveyed via a HTTP header or within a frame.
 
+## Impact on other connections
```
LPardue, line 243:
```suggestion
## Impact on Other Connections
```

draft-kazuho-httpbis-priority.md
```diff
@@ -240,6 +240,18 @@ compatible for peers that are unaware of the extended meaning.
...
+
+When multiple connections are open, the priority MAY have impact on the
```
LPardue, line 245: For my own benefit at least, I would like to understand what exactly "multiple connections" means in this context. There is a difference to multiple HTTP connections to the same authority (something not recommended for H2 or H3) vs multiple connections to different authorities that all compose a page (but are serviced by the same physical machine) vs just the client having multiple TCP connections open .

kazuho (commented):
How about going like:

_A server MUST NOT use the priority values for making schedule decisions across multiple connections, unless it knows that those connections originate from the same client. Due to this, priority information conveyed over a non-coalesced HTTP connection (e.g., HTTP/1.1) might go unused._

_A client MAY use the priority values for making local scheduling choices for the requests it initiates._

The intent here is to provide clearer guidance for implementors by starting with subject (i.e. server or client) and then a RFC 2119 verb.

LPardue:
I agree with @kazuho and think his suggested wording would address the comment that I made.

kazuho:
PS. re "a server MUST NOT use the priority values for making schedule decisions across multiple connections," I am starting to wonder if we need to be that strict. One example that will be prohibited by this provision is the use of a different congestion controller for a "background" request.

I think it might make sense to change this to SHOULD NOT, expand the Security Considerations to four sub-topics: general discussion on fairness (this PR), coalescing intermediaries (existing section 7.1), HTTP/1 (see #107), and use of different congestion controllers.

ianswett:
Feel free to close this once #110 lands if you'd like.

PR #109, opened 2019-11-19
Clarify when PRIORITY_UPDATE can be sent

ianswett:
And what to do if it's received out of order in HTTP/3.

I was going to file an issue, but instead this PR sticks with existing behavior, and tightens up the constraints some.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -474,7 +474,7 @@ Empty:
 : A seven-bit field that has no semantic value.
 
 The HTTP/3 PRIORITY_UPDATE frame MUST NOT be sent with an invalid identifier including before the request stream
-has been opened or a PUSH_PROMISE for the corresponding PUSH_ID has been
+has been opened or a before a promised request has been
 received.
 
 In HTTP/3, the PRIORITY_UPDATE frame can arrive before any data is received
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -485,6 +485,17 @@ be opened, the frame SHOULD be buffered until the stream is opened and
 applied immediately after the HEADERS are parsed.  This consumes extra state
 on the peer, but existing stream limits bound the size of that state.
 
+The HTTP/3 PRIORITY_UPDATE frame is sent on the control stream. Because there
+are no ordering guarantees between streams, a client that reprioritizes a
+request before receiving the response data might cause the server to receive
+a PRIORITY_UPDATE for an unknown request. If the request stream ID is within
+bidirectional stream limits, the PRIORITY_UPDATE frame SHOULD be buffered
+until the stream is opened and applied immediately after the request message
+has been processed. Holding PRIORITY_UPDATES consumes extra state on the peer,
+although the size of the state is bounded by bidirectional stream limits. There
+is no bound on the number of PRIORITY_UPDATES that can be sent, so an
+endpoint SHOULD store only the most recently received frame.
+
 TODO: add more description of how to handle things like receiving
 PRIORITY_UPDATE on wrong stream, a PRIORITY_UPDATE with an invalid ID, etc.
 
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -408,7 +408,7 @@ HTTP/2 this is the Stream ID, in HTTP/3 this is either the Stream ID or Push ID.
 In HTTP/2 and HTTP/3, after a request message is sent on a stream, the stream
 transitions it into a state that prevents the client from sending additional
 frames on the stream. Modifying this behavior would require a semantic change
-to the protocol. this is avoided by restricting the stream on which a
+to the protocol, but this is avoided by restricting the stream on which a
 PRIORITY_UPDATE frame can be sent. In HTTP/2 the frame is on stream zero and
 in HTTP/3 it is sent on the
 control stream ({{!I-D.ietf-quic-http}}, Section 6.2.1).
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -476,7 +476,9 @@ Empty:
 
 The HTTP/3 PRIORITY_UPDATE frame MUST NOT be sent with an invalid identifier,
 including before the request stream has been opened or before a promised
-request has been received.
+request has been received.  If a server receives a PRIORITY_UPDATE specifying
+a push ID that has not been promised, it SHOULD be treated as a connection
+error of type H3_ID_ERROR.
 
 Because the HTTP/3 PRIORITY_UPDATE frame is sent on the control stream and
 there are no ordering guarantees between streams, a client that reprioritizes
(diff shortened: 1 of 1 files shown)


kazuho (approved):
LGTM! Thank you for the PR.

draft-kazuho-httpbis-priority.md
```diff
@@ -434,6 +435,9 @@ frame header MUST be zero (0x0).
 ~~~
 {: #fig-h2-reprioritization-frame title="HTTP/2 PRIORITY_UPDATE Frame Payload"}
 
+The PRIORITY_UPDATE frame in HTTP/2 MUST NOT be sent prior to opening the
+stream.
```
LPardue: I'd like to capture what the outcome of this MUST NOT is. A stream error on stream 0 is pretty catestrophic so do you expect to blow up the connection or just ignore the frame?

I think whatever we pick is a good bucket to drop all the other error cases that might happen (but that we can expand in separate PRs).

draft-kazuho-httpbis-priority.md
```diff
@@ -434,6 +435,9 @@ frame header MUST be zero (0x0).
 ~~~
 {: #fig-h2-reprioritization-frame title="HTTP/2 PRIORITY_UPDATE Frame Payload"}
 
+The PRIORITY_UPDATE frame in HTTP/2 MUST NOT be sent prior to opening the
```
```suggestion
The HTTP/2 PRIORITY_UPDATE frame MUST NOT be sent prior to opening the
```

draft-kazuho-httpbis-priority.md
```diff
@@ -469,6 +473,17 @@ Element ID is interpreted as a Push ID.
 Empty:
 : A seven-bit field that has no semantic value.
 
+The PRIORITY_UPDATE frame in HTTP/3 MUST NOT be sent before the stream
```
```suggestion
The HTTP/3 PRIORITY_UPDATE frame MUST NOT be sent with an invalid identifier including before the request stream
```

draft-kazuho-httpbis-priority.md
```diff
@@ -469,6 +473,17 @@ Element ID is interpreted as a Push ID.
...
+has been opened or a PUSH_PROMISE for the corresponding PUSH_ID has been
```
```suggestion
has been opened or a before a promised request has been
```

draft-kazuho-httpbis-priority.md
```diff
@@ -469,6 +473,17 @@ Element ID is interpreted as a Push ID.
...
+received.
+
+In HTTP/3, the PRIORITY_UPDATE frame can arrive before any data is received
+on the corresponding stream or server push due to reordering or proactively
+changing the priority of a server push.  If the stream ID is one which can
+be opened, the frame SHOULD be buffered until the stream is opened and
+applied immediately after the HEADERS are parsed.  This consumes extra state
+on the peer, but existing stream limits bound the size of that state.
```
LPardue: I think this is _almost_ there in achieving this goal. I think we should split out the error cases because of the difference in the handling of request and push IDs. For example,

```
The HTTP/3 PRIORITY_UPDATE frame is sent on the control stream. Because there are no ordering
guarantees between streams, a client that reprioritizes requests before receiving response data might cause 
the server to receive a PRIORITY_UPDATE for an unknown request. If the request stream ID is within 
bidirectional stream limits, the PRIORITY_UPDATE frame SHOULD be buffered until the stream is opened and 
applied immediately after the request message has been processed. Holding PRIORITY_UPDATES in this way
 consumes extra state on the peer, although the size of state is bounded by bidirectional stream limits there is 
no bound on the number of PRIORITY_UPDATES that a client can send. An endpoint that wishes to minimise
 state might decide to store only the most recently received frame.
```

then we need a paragraph that explain the specifics of the push ID space, which has a few differences.

LPardue (changes requested):

draft-kazuho-httpbis-priority.md
ianswett: I think blowing up is fine.  I'll clarify if this happens, it's ok to close the connection.

draft-kazuho-httpbis-priority.md
```diff
@@ -434,6 +435,10 @@ frame header MUST be zero (0x0).
 ~~~
 {: #fig-h2-reprioritization-frame title="HTTP/2 PRIORITY_UPDATE Frame Payload"}
 
+The HTTP/2 PRIORITY_UPDATE frame MUST NOT be sent prior to opening the
+stream.  If a PRIORITY_UPDATE is received prior to the stream being opened,
+it MAY close the connection error.
```
```suggestion
it MAY be treated as a connection error of type PROTOCOL_ERROR.
```

draft-kazuho-httpbis-priority.md
```diff
@@ -469,6 +474,21 @@ Element ID is interpreted as a Push ID.
...
+a request before receiving the response data might cause the server to receive
+a PRIORITY_UPDATE for an unknown request. If the request stream ID is within
+bidirectional stream limits, the PRIORITY_UPDATE frame SHOULD be buffered
+until the stream is opened and applied immediately after the request message
+has been processed. Holding PRIORITY_UPDATES consumes extra state on the peer,
+although the size of the state is bounded by bidirectional stream limits. There
+is no bound on the number of PRIORITY_UPDATES that can be sent, so an
+endpoint SHOULD store only the most recently received frame.
+
 TODO: add more description of how to handle things like receiving
```
LPardue, line 494: for server push there are different concerns, mainly because the client controls the MAX_PUSH_ID space. Imagine a client that sends MAX_PUSH_ID to MAX_INT and then issues MAX_INT copies of PRIORITY_UPDATE before any requests are sent.

I think we may want to say that the server SHOULD NOT buffer PRIORITY_UPDATES that refer to a push ID that has not been promised by the server.

draft-kazuho-httpbis-priority.md
```diff
@@ -405,12 +405,13 @@ reprioritization. It carries updated priority parameters and references the
...
 HTTP/2 this is the Stream ID, in HTTP/3 this is either the Stream ID or Push ID.
 
-In HTTP/2 and HTTP/3 a request message sent on a stream transitions it into a
-state that prevents the client from sending additional frames on the stream.
-Modifying this behavior requires a semantic change to the protocol, this is
-avoided by restricting the stream on which a PRIORITY_UPDATE frame can be sent.
-In HTTP/2 the frame is on stream zero and in HTTP/3 it is sent on the control
-stream ({{!I-D.ietf-quic-http}}, Section 6.2.1).
+In HTTP/2 and HTTP/3, after a request message is sent on a stream, the stream
+transitions it into a state that prevents the client from sending additional
```
```suggestion
transitions to a state that prevents the client from sending additional
```

LPardue (approved):
approved modulo a nit

ianswett:
Thanks, fix applied.

draft-kazuho-httpbis-priority.md
ianswett: I was thinking if you receive a PRIORITY_UPDATE for a push ID that hasn't been promised, it's a fatal error?

ianswett: I added some text to clarify that you should close the connection if an un-promised Push ID is received.

kazuho:
Thank you! Merged.

PR #111, opened 2019-11-20
reorganize

### LINKED ISSUE #102: Unspecified handling of `urgency` values outside the defined range
The defined range of the urgency parameter is [0,7]. We don't have any text that explains how endpoints generate or consume urgency parameters outside the range.

One option for the endpoints that receive such a parameter is to ignore it. This is problematic if ignoring means use default urgency level.

Another option is to specify that the values outside the range are floored/clamped. e.g. a value less than 0 MUST be interpreted as 0, a value greater 7 MUST be interpreted as 7.

kazuho:
Followup on #104.

Define parameters, then talk about headers and frames. Instead of calling the scheme as extensible, we state that the parameters can be added.

Closes #102.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -226,38 +226,26 @@ extensible priority scheme when and only when all the requests that are to be
 sent on that backend connection originates from one client-side connection that
 has negotiated the use of the extensible priority scheme (see {{fairness}}).
 
-# Extensible Priorities
+# The Priority Parameters
 
-This priority design is intended to be extensible, and uses a sequence of
-key-value pairs encoded as a Structured Headers Dictionary
-({{!STRUCTURED-HEADERS}}) to enable extensibility. Each dictionary member
-represents a parameter of the Priority header field.
+The priority information is a sequence of key-value pairs, providing room for
+future extensions. Each key-value pair represents a priority parameter.
 
-When attempting to extend priorities, care must be taken to ensure any use of
-existing parameters are either unchanged or modified in a way that is backwards
-compatible for peers that are unaware of the extended meaning.
+The Priority HTTP header field is used to transmit this set of parameters when
+a request or a response is being issued. In order to request reprioritization
+after the requests have been issued, HTTP-version-specific frames are used by
+clients to transmit this same information.
 
-The scheme has a single encoding and set of functionality whether it's
-conveyed via a HTTP header or within a frame.
+In both cases, the set of priority parameters is encoded as a Structured Headers
+Dictionary ({{!STRUCTURED-HEADERS}}).
 
-# The Priority HTTP Header Field
+This document defines the urgency(`u`) and incremental(`i`) parameters. When
+used, these parameters MUST be accompanied by values. When any of the defined
+parameters are omitted, or if the Priority header field is not used, their
+default values SHOULD be applied.
 
-The Priority HTTP header field can appear in requests and responses. A client
-uses it to specify the priority of the response. A server uses it to inform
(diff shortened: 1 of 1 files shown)


kazuho:
@ianswett @LPardue Does this look alright?

draft-kazuho-httpbis-priority.md
```diff
@@ -226,38 +226,26 @@ extensible priority scheme when and only when all the requests that are to be
 sent on that backend connection originates from one client-side connection that
 has negotiated the use of the extensible priority scheme (see {{fairness}}).
 
-# Extensible Priorities
+# The Priority Parameters
```
ianswett, line 229:
```suggestion
# Priority Parameters
```

draft-kazuho-httpbis-priority.md
```diff
@@ -226,38 +226,26 @@ extensible priority scheme when and only when all the requests that are to be
...
-conveyed via a HTTP header or within a frame.
-
-# The Priority HTTP Header Field
-
-The Priority HTTP header field can appear in requests and responses. A client
-uses it to specify the priority of the response. A server uses it to inform
-the client that the priority was overwritten. An intermediary can use the
-Priority information from client requests and server responses to correct or
-amend the precedence to suit it (see {{merging}}).
+The Priority HTTP header field is used to transmit this set of parameters when
```
ianswett, line 234:
```suggestion
The Priority HTTP header field is an end-to-end way to transmit this set of parameters when
```

draft-kazuho-httpbis-priority.md
```diff
@@ -226,38 +226,26 @@ extensible priority scheme when and only when all the requests that are to be
...
+a request or a response is being issued. In order to request reprioritization
+after the requests have been issued, HTTP-version-specific frames are used by
```
ianswett, line 236:
```suggestion
after a request have been issued, HTTP-version-specific frames are used by
```

draft-kazuho-httpbis-priority.md
```diff
@@ -226,38 +226,26 @@ extensible priority scheme when and only when all the requests that are to be
...
+clients to transmit this same information.
```
ianswett, line 237:
```suggestion
clients to transmit the same information on a single hop.
```

ianswett: Related Q: Do we want to allow the server to send the frame?

kazuho:
Sorry folks, but let me merge this as I think this would unblock other changes from coming in.

draft-kazuho-httpbis-priority.md
```diff
@@ -384,6 +372,23 @@ set to `4` and the incremental parameter set to `1`.
...
+
+# The Priority HTTP Header Field
+
+The Priority HTTP header field can appear in requests and responses. A client
+uses it to specify the priority of the response. A server uses it to inform
+the client that the priority was overwritten. An intermediary can use the
+Priority information from client requests and server responses to correct or
+amend the precedence to suit it (see {{merging}}).
+
+The Priority header field is an end-to-end signal of the request
```
ianswett, line 389: Should we specifically say that the intermediary should not change the priority and if it wants to apply a different priority on the next hop, it should use a frame?  I feel like we get confused when we start talking about intermediaries modifying the priority header in various ways.

ianswett (commented):
Some small suggestions, but thanks for doing this reorg.  It's a huge improvement.

ianswett:
I'm creating a new PR with my suggestions, so you can ignore them until that's sent out.

LPardue:
didn't get a chance to review but I support the drive for progress

PR #113, opened 2019-11-20
considerations on caching behavior

### LINKED ISSUE #97: Talk about caching issues stemming from HEADERS-based approach
Discussed in Singapore. Caching-issues related to header-based approach are difficult, but part of "normal HTTP". 

We should provide guidance though and list the foreseen gotcha's to prevent people from making obvious mistakes.

I don't have enough experience to know what those gotcha's actually are, but e.g., stripping header when storing in cache, interaction with Vary, etc. are things I've heard multiple times. cc @ianswett

kazuho:
closes #97.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -386,8 +386,15 @@ the client that the priority was overwritten. An intermediary can use the
 Priority information from client requests and server responses to correct or
 amend the precedence to suit it (see {{merging}}).
 
-The Priority header field is an end-to-end signal of the request
-priority from the client or the response priority from the server.
+The Priority header field is an end-to-end signal of the request priority from
+the client or the response priority from the server.
+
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -615,6 +615,15 @@ of delaying the delivery of updates.
 Also, a client MAY use the priority values for making local scheduling choices
 for the requests it initiates.
 
+## Caching of the Priority Response Header
```
ianswett: Is it worth referencing RFC7231 and RFC7234?

Also, I'd prefer to see this in the section about the header, now that that section is much shorter.

kazuho: It references RFC 7234. I've moved the paragraph to the header section, I agree that it would be more natural to explain what the header being end-to-end means, when pointing out that the header is end-to-end.

ianswett (approved):

PR #118, opened 2019-11-21
Remove outdated Considerations

kazuho:
We now provide a solution in Section 3: _If intermediaries want to specify prioritizaton on a multiplexed HTTP connection, it SHOULD use a PRIORITY_UPDATE frame and SHOULD NOT change the Priority header field._


LPardue (approved):
lgtm, really helps to avoid using an example header that we don't actually define

PR #119, opened 2019-11-21
response MUST NOT be affected by the Priority request header

kazuho:
Forbid servers using the Priority request header to generate different responses, as suggested by @royfielding.

### DIFF
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -335,7 +335,7 @@ The Priority header field is an end-to-end signal of the request priority from
 the client or the response priority from the server.
 
 The sole intention of prioritization is to optimize how responses are being
-provided, not to change what is provided as response. Therefore, a server MUST
+provided, not to change what is provided as the response. Therefore, a server MUST
 NOT generate different HTTP response based on the value of the Priority request
 header field. Sending a cacheable response that depends on the Priority request
 header field might lead to mis-prioritization, when a cache reuses that
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -336,7 +336,7 @@ the client or the response priority from the server.
 
 The sole intention of prioritization is to optimize how responses are being
 provided, not to change what is provided as the response. Therefore, a server MUST
-NOT generate different HTTP response based on the value of the Priority request
+NOT generate a different HTTP response based on the value of the Priority request
 header field. Sending a cacheable response that depends on the Priority request
 header field might lead to mis-prioritization, when a cache reuses that
 response.
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -336,7 +336,7 @@ the client or the response priority from the server.
 
 The sole intention of prioritization is to optimize how responses are being
 provided, not to change what is provided as response. Therefore, a server MUST
-NOT generate different response based on the value of the Priority request
+NOT generate different HTTP response based on the value of the Priority request
 header field. Sending a cacheable response that depends on the Priority request
 header field might lead to mis-prioritization, when a cache reuses that
 response.
--- a/draft-kazuho-httpbis-priority.md
+++ b/draft-kazuho-httpbis-priority.md
@@ -334,9 +334,12 @@ amend the precedence to suit it (see {{merging}}).
 The Priority header field is an end-to-end signal of the request priority from
 the client or the response priority from the server.
 
-A server MUST NOT generate different response based on the value of the Priority
-request header field. The sole intention of prioritization is to optimize how
-responses are being provided, not to change what is provided as response.
+The sole intention of prioritization is to optimize how responses are being
+provided, not to change what is provided as response. Therefore, a server MUST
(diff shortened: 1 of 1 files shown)


draft-kazuho-httpbis-priority.md
```diff
@@ -334,12 +334,9 @@ amend the precedence to suit it (see {{merging}}).
 The Priority header field is an end-to-end signal of the request priority from
 the client or the response priority from the server.
 
-As is the ordinary case for HTTP caching ({{?RFC7234}}), a response with a
-Priority header field might be cached and re-used for subsequent requests.
-When an origin server generates the Priority response header field based on
-properties of an HTTP request it receives, the server is expected to control the
-cacheability or the applicability of the cached response, by using header fields
-that control the caching behavior (e.g., Cache-Control, Vary).
+A server MUST NOT generate different response based on the value of the Priority
```
ianswett: It can generate a different Priority header in response, correct?

LPardue (approved):
lgtm

rmarx:
I'm afraid we're losing a bit of guidance on the caching footguns with this? This might be clear to you all, but not to me and probably not to some other eventual implementers either. Since this was a main point from Ian's, I wonder if we need to add some more guidance or details in here (not necessarily mentioning Vary etc.)

ianswett:
I agree with @rmarx  that we shouldn't remove the other paragraph and only replace it with this.

draft-kazuho-httpbis-priority.md
kazuho: In 9fffd45, I've clarified that we are talking about "HTTP response" but not a response header field. Does that look better?

kazuho:
@rmarx @ianswett I can see the concern, therefore added a sentence explaining the negative consequence that people might face if they do not obey to the requirement. PTAL.

draft-kazuho-httpbis-priority.md
```diff
@@ -334,12 +334,12 @@ amend the precedence to suit it (see {{merging}}).
 The Priority header field is an end-to-end signal of the request priority from
 the client or the response priority from the server.
 
-As is the ordinary case for HTTP caching ({{?RFC7234}}), a response with a
-Priority header field might be cached and re-used for subsequent requests.
-When an origin server generates the Priority response header field based on
-properties of an HTTP request it receives, the server is expected to control the
-cacheability or the applicability of the cached response, by using header fields
-that control the caching behavior (e.g., Cache-Control, Vary).
+The sole intention of prioritization is to optimize how responses are being
```
ianswett, line 337: I'm not sure it's the sole intention of prioritization.  Don't we say you can use this to inform request behavior as well?

LPardue: I think we are trying to say that the intention of prioritization is to optimise HOW responses are being provided, not WHAT is provided.  I started to walk this path in the introduction but didn't want to get too much into HTTP-core leet speak. IIUC what we are trying to say, I think we can riff off the following text fragments from the intro:

```

   To provide meaningful representation of a document at the earliest
   moment, it is important for an HTTP server to prioritize the HTTP
   responses, or the chunks of those HTTP responses, that it sends.
```

``` 
   This document defines the Priority HTTP header field that can be used
   by both client and server to specify the precedence of HTTP responses
```

draft-kazuho-httpbis-priority.md
```diff
@@ -334,12 +334,12 @@ amend the precedence to suit it (see {{merging}}).
...
+provided, not to change what is provided as response. Therefore, a server MUST
```
ianswett:
```suggestion
provided, not to change what is provided as the response. Therefore, a server MUST
```

draft-kazuho-httpbis-priority.md
```diff
@@ -334,12 +334,12 @@ amend the precedence to suit it (see {{merging}}).
...
+NOT generate different HTTP response based on the value of the Priority request
```
ianswett:
```suggestion
NOT generate a different HTTP response based on the value of the Priority request
```

ianswett (approved):
Some nits, but I think this LG now.

LPardue (approved):

draft-kazuho-httpbis-priority.md
```diff
@@ -334,12 +334,12 @@ amend the precedence to suit it (see {{merging}}).
...
-As is the ordinary case for HTTP caching ({{?RFC7234}}), a response with a
-Priority header field might be cached and re-used for subsequent requests.
-When an origin server generates the Priority response header field based on
-properties of an HTTP request it receives, the server is expected to control the
-cacheability or the applicability of the cached response, by using header fields
-that control the caching behavior (e.g., Cache-Control, Vary).
+The sole intention of prioritization is to optimize how responses are being
+provided, not to change what is provided as the response. Therefore, a server MUST
+NOT generate a different HTTP response based on the value of the Priority request
+header field. Sending a cacheable response that depends on the Priority request
```
ianswett, line 340: I'm finding this last sentence somewhat vague, how about:
"Sending a cacheable response including a Priority response header field that depends upon the Priority request header field is likely to lead to mis-prioritization when a cache reuses that response."

ianswett (commented):
One suggestion

